PR #55, opened 2022-08-11
MADR 0001: Reorganize Documentation

m-ildefons:
Propose new organization for the documentation. The purpose of
reorganizing the documentation is to make it easier for people
unfamiliar with the project to find relevant information.
In order to do that a new hirarchy is proposed. The hirarchy
specifically has the needs of users who are not contributing, users who
are contributing feedback and then code contributors in that order in
mind.

### DIFF
--- a/docs/decisions/0001-reorganize-documentation.md
+++ b/docs/decisions/0001-reorganize-documentation.md
@@ -1,0 +1,59 @@
+---
+status: proposed
+---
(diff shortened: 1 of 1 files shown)


jhmarina (approved):
I totally agree with the proposed changes. I was in fact preparing a PR that reflects some of them.

PR #61, opened 2022-08-12
ci: add release workflow

m-ildefons:
Add release workflow. The release workflow builds heavily onto the
respective build workflows of the subrepositories. It triggers on tags
and uses the refs of the subrepos and the tag to create artifacts with
the correct content.

This PR is not feature complete, but serves as a starting point to get the different pipelines up and running. As work continues on the release pipelines, more steps will be completed.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,63 @@
+---
+name: Release S3GW
+on:
+  schedule:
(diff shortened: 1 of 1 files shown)


jecluis (approved):
I don't fully understand the whole picture, but am curious to see where this ends up. Thanks for handling this!

PR #62, opened 2022-08-12
ci: run pre-commit checks as part of linting

m-ildefons:
Run pre-commit checks as part of linting workflow.

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -1,16 +1,20 @@
+---
 name: Lint
-on: [push, pull_request, workflow_dispatch]
+on:
+  - push
+  - pull_request
+  - workflow_dispatch
+
 jobs:
   lint:
     runs-on: ubuntu-latest
     steps:
+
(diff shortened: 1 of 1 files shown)


irq0:
Running pre commit duplicates markdown lint and codespell. Would it make sense to remove those?

The pre commit check seems to be a bit slow at 37s. Probably mostly due to downloading and installing it. Is there a way to cache the install somehow?

m-ildefons:
Totally sensible to remove those.

The action is quite slow, but I have no idea if and how caching works for that. The upstream recommendation is to use their CI service https://pre-commit.ci/ which apparently is free for open source projects.
I don't know if it would be a smart move to migrate there.
That being said, it ran much faster without the other steps in place. I think I can live with anything below one minute anyways. I think it's fine for now. We can revisit this if it turns out to be too annoying, but don't think it's worth it to sink much more time into speeding it up just to gain a few seconds.

irq0 (approved):
:+1:

PR #67, opened 2022-08-18
ci: fold release workflow into single job

m-ildefons:
Fold the release workflow into a single multi-step job. This is done by
re-using the checked-out version of the subrepos instead of calling the
respective repo's release workflow on the @main branch. Thereby changes
to the respective subrepo don't break the release pipeline of the
overall project


jhmarina (approved):
lgtm

PR #72, opened 2022-08-22
ci: fix syntax in release workflow

m-ildefons:
The previous change folding the workflow calls into build steps didn't work because:

- Workflow calls have to be jobs and not steps in jobs
- Workflow calls have to name a git ref for the foreign repo to checkout


jhmarina (approved):
lgtm @m-ildefons

PR #84, opened 2022-08-30
feature: add docker-compose file

### LINKED ISSUE #74: chart: add docker compose file
Add a docker compose file. While not the primary use case, docker compose has some distinct advantages that are worth leveraging:
- lowers the bar of entry, especially for people using a mac
- lowers our own setup overhead
- makes it easy for prospecting users
- increases our compatibility with rancher desktop

m-ildefons:
Add docker-compose file for easy local deployment of the s3gw and the ui
with minimal runtime requirements.

Requires docker and docker-compose. Should be compatible with docker
swarm, but has not been tested.

This allows users without access to a kubernetes cluster to quickly
deploy the s3gw including ui by simply executing:

```
docker-compose up
```

fixes: #74

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -1,0 +1,41 @@
+---
+version: "3.4"
+services:
+  s3gw:
+    image: ghcr.io/aquarist-labs/s3gw:latest
+    restart: always
+    command:
+      - "--rgw-backend-store"
+      - "sfs"
+    ports:
+      - "7480:7480"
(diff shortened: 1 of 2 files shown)


votdev (approved):

votdev:
This also works with podman-compose
```
$ sudo sysctl net.ipv4.ip_unprivileged_port_start=80
$ podman-compose up
```

PR #101, opened 2022-09-07
ci: pass credentials to s3gw-tools workflow

m-ildefons:
Pass dockerhub credentials to s3gw-tools workflow during workflow call.


jhmarina (approved):
lgtm

PR #112, opened 2022-09-15
ci: release workflow: use github token

m-ildefons:
Use github.token to authenticate to ghcr.io instead of a dedicated access token/username. This removes the requirement to explicitly configure access credentials and is the preferred method for organizations and shared repositories.


jhmarina (approved):
lgtm

PR #274, opened 2022-12-13
docs: add updated roadmap

jhmarina:
# Describe your changes
This PR updates the product roadmap with the items defined on the internal Confluence [project page](https://confluence.suse.com/display/RES/S3+gateway+project).

### DIFF
--- a/docs/roadmap.md
+++ b/docs/roadmap.md
@@ -1,38 +1,120 @@
+<!-- markdownlint-disable MD024 -->
+
 # Roadmap
 
-The aim is to deliver a Minimum Viable Product (MVP). In a nutshell, an MVP
-seeks to collect the maximum amount of validated learning about our users in a
-short time.
+## 2022
 
-Based on the user feedback we collect, we develop features that add up towards
-the goal of delivering the MVP. We're working on phases:
+### Quarter 1
 
-## Phase 1
-
-- Use SQLite as storage backend.
-- Running the s3gw in a container.
-- Be able to run basic S3 operations.
-- Ability to deploy the S3 Gateway on a K8s or K3s cluster with.
-- [Design mockups][1] for User management & an S3 explorer.
+- Brainstorm process
 
 ---
 
-## Phase 2
+### Quarter 2
 
-- Basic S3 explorer UI
-- Helm chart
-- Define & start work on a file-based backend
-- Implement CI/CD
-
----
-
-## Phase 3
-
-- File-based backend integration
-- Increased support for S3 functional features
-- S3 explorer UI
-- User management UI
+- Initial project setup
+- First tests with RGW
+- [UI mockups][1]
 
 ---
 
-[1]: https://www.figma.com/file/IeozuvvYlrKBs7qm030dyo/S3-Wireframe?node-id=0%3A1
+### Quarter 3
+
+- File-based backend
+- Basic S3 operations support
+- User management
+- Helm charts
+- Testing
+- UI implementation
+- Documentation
+- Basic object management & versioning
+- Bucket management
+- Automation implementation
+
+---
+
+### Quarter 4
+
+- Object deletion/undeletion
+- Multipart uploads
+- Multipart copy
+- ACL support
+- UI: Basic object explorer
+- Rancher Partner chart
+
+## 2023
+
+### Plan of record
+
+Features the team is committed to deliver within the timeframe.
+
+#### Quarter 1
+
+- [Expiration Lifecycle Management (unversioned)][2]
+- [Object locks][3]
+- [Object listing with prefixes/filters][4]
+- [Telemetry][5]
+- [UI: Object deletion/undeletion][7]
+
+---
+
+#### Quarter 2
+
+- [Expiration Lifecycle Management (versioned)][8]
+- [Access & Identity Management][10]
+- [Metrics from backend][9]
+- [UI: Object versioning][11]
+
+---
+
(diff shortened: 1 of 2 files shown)


docs/roadmap.md
```diff
@@ -1,38 +1,98 @@
...
 
-## Phase 3
+### Quarter 3
 
-- File-based backend integration
-- Increased support for S3 functional features
-- S3 explorer UI
-- User management UI
+- [Implement Amazon Security Token][12]
+- [Cross-site replication][13]
```
jecluis: I still think this is a bad idea, having it on the roadmap for next year. This should be a "we'll work on it should there be real need, but otherwise it will be pushed further into the future" kind of thing.

docs/roadmap.md
```diff
@@ -1,38 +1,98 @@
...
 
 ---
 
-[1]: https://www.figma.com/file/IeozuvvYlrKBs7qm030dyo/S3-Wireframe?node-id=0%3A1
+### Quarter 4
+
+- [Implement Storage classes][14]
```
jecluis: This, the same as above.  Lets try to push this somewhere else. Maybe a section on "what we would like to see in the future, but not currently planned". It's fine if we don't have anything quite set for Q4 2023, I think. I'm sure things will change as we move through 2023 and things will become clearer.

jecluis (changes requested):

docs/roadmap.md
l-mb: Agreed; we should differentiate between a "plan of record" (i.e., something we're actively committing to achieving within a specific timeframe), "plan of direction" (which where the replication would fit) that we want to get to eventually, "in scope" (something we'd accept patches for, but have no plans for ourselves), and finally "out of scope" (we really don't want to go there unless core assumptions change, say, scaling this to 1000s of nodes).

giubacc: IMO, this is not necessarily a bad idea; a roadmap is by-design less accurate and possibly wrong the more you move the pointer into the future.
This simply represents an intention to do things in the future; but it is known that this is subjected to changes when the future become the present.
There is no need that a thing is an "actual need" to put it on a plan. There are things that we could simply want to do regardless if someone will use them.

jecluis: Keep in mind that the roadmap is the closest we have to an official commitment of intent.

Once it goes on the roadmap we are sending the message this is something we consider to be in scope, regardless of whether it's in Q4 or Q _N_ 202X. And I don't think we want to consider this in scope for now. Hence why it should not be in the roadmap, but which we may still keep somewhere as a "nice to have, but we're not working on it" kind of thing.

jhmarina: Acknowledged. This is something we'd discussed in the past but isn't reflected here. Let me redo the list a bit

jhmarina: Ok, I restructured it, using @l-mb 's suggestion. I think it is best to avoid creating false expectations. I've also improved the disclaimer stating that this roadmap is subject to change.

docs/roadmap.md
jhmarina: changed

jecluis (approved):

docs/roadmap.md
```diff
@@ -1,38 +1,120 @@
...
 ---
 
-[1]: https://www.figma.com/file/IeozuvvYlrKBs7qm030dyo/S3-Wireframe?node-id=0%3A1
+### Plan of direction
+
+Features the team wants to develop but are currently not a top priority.
+
+- [Cross-site replication][13]
+- [Implement Storage classes][14]
+- [CRD based configuration of buckets and rbac][6]
```
m-ildefons, line 88: Abbreviations in capital letters: RBAC

docs/roadmap.md
```diff
@@ -1,38 +1,120 @@
...
+- [SSO][16]
+
+### Out of scope
+
+Features that aren't being considered unless core assumptions change.
+
+!!! Info
+    Please note that this roadmap is an evergreen document and will most
+    certainly evolve as we continue to learn from our users.
+
```
m-ildefons, line 104: A divider here would be nice to indicate that the footnotes aren't the features we consider out of scope.

docs/roadmap.md
```diff
@@ -1,38 +1,120 @@
...
+### Plan of record
 
-- Use SQLite as storage backend.
-- Running the s3gw in a container.
-- Be able to run basic S3 operations.
-- Ability to deploy the S3 Gateway on a K8s or K3s cluster with.
-- [Design mockups][1] for User management & an S3 explorer.
+Features the team is committed to deliver within the timeframe.
+
+#### Quarter 1
```
m-ildefons, line 51: Why are the `Quarter x` headings level 4 in 2023 and level 3 in 2022?

docs/roadmap.md
```diff
@@ -1,38 +1,120 @@
...
+- ACL support
+- UI: Basic object explorer
+- Rancher Partner chart
 
-Based on the user feedback we collect, we develop features that add up towards
-the goal of delivering the MVP. We're working on phases:
+## 2023
 
-## Phase 1
+### Plan of record
```
m-ildefons, line 47: I have no idea what `Plan of record` means in this context. It does seem to apply to the 2022 bullet points too though, so maybe just put it as a general paragraph at the top.

m-ildefons (changes requested):
A bit nitpicky, but the fixes should be quick and easy

PR #308, opened 2023-01-10
ci: add nightly build

m-ildefons:
- Add build workflow for nightly container builds from latest HEAD of aquarist-labs/ceph@s3gw branch.
- Push to quay.io and with nightly-$(date) tag and update floating nightly-latest tag

### DIFF
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -1,0 +1,156 @@
+---
+name: Nightly Builds
+
+on:
+  schedule:
+    cron: '0 0 * * *'  # Every night midnight
+
+  workflow_dispatch:   # Or manually - for testing
+
+jobs:
+  # Build the build-environment container, using it's workflow
+  build-env:
+    runs-on: ubuntu-latest
+    needs:
+      - set-git-refs
+
+    outputs:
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly.yaml
```diff
@@ -0,0 +1,156 @@
...
+        run: |
+          DATE="$(date +%Y-%m-%d)"
+          echo "date=${DATE}" >> $GITHUB_OUTPUT
+
+      - name: Build Buildenv Container
+        uses: docker/build-push-action@v3
+        with:
+          push: true
+          tags: ${{ secrets.DOCKERHUB_USERNAME }}/buildenv-radosgw:nightly-${{ steps.date.outputs.date }}
+          file: tools/build/Dockerfile.build-radosgw
```
jecluis, line 50: can't we rely on the containers we have on quay?

jecluis (commented):
lgtm, and it's good to go once the one question I have is answered :)

.github/workflows/nightly.yaml
m-ildefons: Good point. Copied that code from the release pipeline with minimal modifications, so this is slipped past.

jecluis (approved):

PR #336, opened 2023-02-09
ci: build-environment triggers and fixes

m-ildefons:
- Use docker actions with login to ensure authentication succeeds for the push of the container images to quay.io
- Allow manually triggering the workflow
- Build the environments only on changes to the respective source files otherwise


jhmarina (approved):
lgtm

PR #348, opened 2023-02-17
ci: on-disk format checker and release tests

m-ildefons:
- Add on-disk format checker able to detect the most egregious breaking on-disk format changes.
- Incorporate simple tests into release workflow
- Remove usage of Dockerhub from release workflow in favor of Quay.io
- Fix whitespace errors

The purpose of the on-disk format change test is to ensure a release with on-disk format changes has an appropriate notice in the release notes.
The purpose of the smoke test is to ensure the packaging into the container image has been successful and the runtime works as expected.

### DIFF
--- a/tools/tests/on-disk-format-checker.sh
+++ b/tools/tests/on-disk-format-checker.sh
@@ -1,0 +1,157 @@
+#!/bin/bash
+#
+# Copyright 2022 SUSE, LLC.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# ------
+#
+# This script performs tests to ensure that no unintended breaking changes are
+# introduced to the on-disk format of the s3gw.
+# It works by spinning up an old version of the s3gw, uploading a couple objects
+# to it, then reusing that volume with a new version of the s3gw, trying to read
+# those objects again and ensuring integrity with checksums along the way.
+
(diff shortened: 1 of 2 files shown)


tools/tests/on-disk-format-checker.sh
```diff
@@ -0,0 +1,164 @@
...
+  sha256sum ./* > checksums
+  popd > /dev/null || exit 1
+}
+
+_cleanup() {
+  local result="$?"
+
+  _stop_s3gw
+  debug "Cleaning up..."
+	rm -rf "$VOL"
```
irq0: Why is this indented?

tools/tests/on-disk-format-checker.sh
```diff
@@ -0,0 +1,164 @@
+#!/bin/bash
```
irq0, line 1: lgtm. 


Some nits:

Calling cleanup using an EXIT handler may be simpler and save a couple of calls to _cleanup.

Example:
```
trap cleanup EXIT
function cleanup()
{

}
```

Consider removing the "_" function prefix. It is a pythonism for "private" fuctions, not? In this small script it IMHO doesn't add much value to the reader.


Consider set -x instead of the debug print logic.

Using set -e and if command or if ! command could be clearer for the reader to understand what may or may not fail.

irq0 (dismissed):

tools/tests/on-disk-format-checker.sh
m-ildefons: Accidental. :facepalm: I tabbed when I should have spaced and by pure coincidence my editor shows this with the same indentation in this case.

tools/tests/on-disk-format-checker.sh
m-ildefons: I am not a fan of `set -x` most of the time, especially not as default since it makes the output hard to read.
The exit handler is a good call though.

The thing with the `_` prefixes is something I like to do to a) signify that should someone source this file these functions aren't meant to be used willy-nilly and b) to avoid naming conflicts with stuff on `$PATH`. I also makes it quite clear that these aren't some command from somewhere but functions in this script. So in a sense these _are_ private functions. But I'll remove them if you think this is detrimental to readability.

irq0 (approved):
lgtm

PR #356, opened 2023-02-24
ci: s3tests runner log from container

m-ildefons:
Make the s3tests runner script collect logs, when running the s3gw within the runtime container. Previously, logs of the s3gw would only be collected when running in the host environment. However this isn't how the tests are run in the CI system, so in order to be able to collect logs there, the container runtime has to be instructed to write out its logs to a file, from where the logs can be collected.


jecluis (dismissed):

jecluis (dismissed):
lgtm

jecluis (approved):

PR #379, opened 2023-03-07
ci: fix on disk format checker on container crash

0xavi0:
When testing the new container for on-disk format that container could crash or exits because `rgw` detects that the metadata format has changed.

In those cases this PR checks for the logs, prints them to stdout and returns 1.


m-ildefons (approved):
LGTM

PR #392, opened 2023-03-10
ci: unify container build

m-ildefons:
Unify the container build by creating a multi-stage contaienr which takes a copy of the ceph subrepo and builds the radosgw binary with the sfs backend inside it. The next stage then creates a runtime container capable of executing the resulting artefacts.
The total result of this is that building the s3gw container is no more compliacted that a `docker build -t s3gw .`

Requires-pr: aquarist-labs/ceph#130

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,6 +5,36 @@ RUN zypper ar \
   s3gw-deps \
  && zypper --gpg-auto-import-keys ref
 
+RUN zypper -n install \
+  libblkid1 \
+  libexpat1 \
+  libtcmalloc4 \
+  libfmt9 \
+  liboath0 \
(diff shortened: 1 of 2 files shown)


Dockerfile
jecluis, line 1: should this file be in the root of the repo? Also, doesn't this supersede other Dockerfiles in the repo? Shouldn't those be removed?

m-ildefons: Being in the root of the repo in this case just simplifies the necessary build commands. As it stands now, it does not deprecate the other Dockerfiles in `tools/build`, because the container images they produce are still in use with the ci pipeline in aquarist-labs/ceph for running tests on PRs

jecluis (approved):

PR #418, opened 2023-03-29
Add minikube instant demo script

irq0:
Depends on https://github.com/aquarist-labs/s3gw-charts/pull/94

Run ./instant-demo.sh. Get minikube with S3GW, S3GW-UI and Traefik
Ingress controller in a couple of minutes.

Defaults to a non-SSL setup. Set USE_SSL=true to enable
a cert manager based SSL setup.

Signed-off-by: Marcel Lauhoff <marcel.lauhoff@suse.com>

# Describe your changes

### DIFF
--- a/tools/minikube/instant-demo.sh
+++ b/tools/minikube/instant-demo.sh
@@ -1,0 +1,78 @@
+#!/bin/bash
+set -e
+set -u
+
+: "${USE_SSL:=false}"
+: "${S3GW_CHARTS:="https://aquarist-labs.github.io/s3gw-charts/"}"
+
+if ! [[ $S3GW_CHARTS =~ https?:// || -r "${S3GW_CHARTS}/Chart.yaml" ]]; then
+	echo "S3GW_CHARTS set to something weird ($S3GW_CHARTS)"
+	echo "Please set to the s3gw charts dir (e.g s3gw-charts/charts/s3gw) or a remote location"
+	exit 1
+fi
+
+set -x
+
+minikube start --driver=kvm2 --extra-disks=0
+
+helm repo add jetstack https://charts.jetstack.io
+helm repo add traefik https://traefik.github.io/charts
+
+helm repo update
+
+backend_proto="http"
+
+if [[ $USE_SSL == "true" ]]; then
+	backend_proto="https"
+	helm install cert-manager \
+		--namespace cert-manager jetstack/cert-manager \
+		--create-namespace \
+		--set installCRDs=true \
+		--set extraArgs[0]=--enable-certificate-owner-ref=true
+fi
+helm install traefik-ingress traefik/traefik
+
+traefik_cluster_ip="$(kubectl \
+	get service \
+	traefik-ingress -ojsonpath='{.spec.clusterIP}')"
+helm_values_file="$(mktemp)"
+trap "rm $helm_values_file" EXIT
+
+cat >"$helm_values_file" <<EOF
+ingress:
+  enabled: true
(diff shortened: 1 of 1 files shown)


tools/minikube/instant-demo.sh
```diff
@@ -0,0 +1,75 @@
...
+
+: "${USE_SSL:=false}"
+: "${backend_proto:="http"}"
+: "${S3GW_CHARTS:="$(dirname "$0")/../../charts/charts/s3gw"}"
+
+if [[ ! -r "${S3GW_CHARTS}/Chart.yaml" ]]; then
+    echo "S3GW_CHARTS set to something weird ($S3GW_CHARTS)"
+    echo "Please set to the s3gw charts dir (e.g s3gw-charts/charts/s3gw)"
+    exit 1
+fi
```
m-ildefons: Instead of bailing when no local charts repo is present, why not do
`helm repo add s3gw https://aquarist-labs.github.io/s3gw-charts` and install an official release?

tools/minikube/instant-demo.sh
```diff
@@ -0,0 +1,75 @@
+#!/bin/bash
+set -e
+set -u
+
+: "${USE_SSL:=false}"
+: "${backend_proto:="http"}"
```
m-ildefons: This shouldn't be set by a user ever. It is depending on `USE_SSL`.

tools/minikube/instant-demo.sh
```diff
@@ -0,0 +1,75 @@
...
+tlsIssuer: "s3gw-issuer"
+email: "mail@example.com"
+ui:
+  enabled: true
+  serviceName: "s3gw-ui"
+  publicDomain: "fe.${traefik_cluster_ip}.omg.howdoi.website"
+  backendProtocol: "${backend_proto}"
+serviceName: "s3gw"
+useExistingSecret: false
+defaultUserCredentialsSecret: "s3gw-creds"
```
m-ildefons, line 55: When `useExistringSecret` is `false`, there is no need to set `defaultUserCredentialSecret`. It doesn't hurt, so I wouldn't block this PR on it, but just FYI.

tools/minikube/instant-demo.sh
```diff
@@ -0,0 +1,75 @@
...
+	 --set installCRDs=true \
+	 --set extraArgs[0]=--enable-certificate-owner-ref=true
+fi
+helm install traefik-ingress traefik/traefik
+
+traefik_cluster_ip="$(kubectl \
+			      get service \
+			      traefik-ingress -ojsonpath='{.spec.clusterIP}')"
+
+cat > /tmp/s3gw-helm-values.yaml <<EOF
```
m-ildefons: Please add a cleanup function so this script doesn't litter artifacts all over the place.

m-ildefons (changes requested):

tools/minikube/instant-demo.sh
irq0: :+1:

tools/minikube/instant-demo.sh
irq0: Good idea! It's an even better default

tools/minikube/instant-demo.sh
irq0: ack

tools/minikube/instant-demo.sh
irq0: :+1:

m-ildefons (approved):
LGTM

PR #430, opened 2023-04-05
ci/s3tests: add extra parameter added for lifecycle s3tests

0xavi0:
We need to add an extra parameter in order to run the lifecycle tests.

Once this is done we can add the lifecycle s3tests to the list of tests to be executed.


m-ildefons (approved):
LGTM

PR #434, opened 2023-04-12
ci: run s3tests in parallel

m-ildefons:
Speed up s3tests by running them in parallel.
Each test is running against its own instance of the s3gw anyways, since this is needed to limit the blast radius of crashes. Thus mutiple tests can be executed concurrently as they do not depend on each other and do not interfere with each other on different instances of the s3gw. GNU parallel provides the job queue and flock is used to synchronize result collection.

### DIFF
--- a/tools/tests/s3tests-runner.sh
+++ b/tools/tests/s3tests-runner.sh
@@ -54,6 +54,8 @@ JOB=
 TMPFILE=
 TMPDIR=
 
+NPROC=${NPROC:-"$(nproc --ignore=2)"}  # Used to run tests in parallel
+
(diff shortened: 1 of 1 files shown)


m-ildefons:
Just as a point of reference, parallel execution of the tests reduced the runtime of the known-to-pass test suite (the stuff that we run for each PR) from ~15 min to ~1:30 min on my machine (sample-size 1).

jecluis (approved):
lgtm

I was initially confused how the `TMPDIR` value would not end up clashing between runs, but then at the end I saw and learnt about `parallel`. :wink:

PR #475, opened 2023-04-25
ci/s3tests-runner.sh : export CONTAINER_EXTRA_PARAMS

### LINKED ISSUE #474: [BUG] s3tests-runner.sh is not exporting CONTAINER_EXTRA_PARAMS
After the s3tests-runner.sh is running s3tests in parallel the `CONTAINER_EXTRA_PARAMS` is no longer exported.

0xavi0:
Lifecycle tests are not working because the `CONTAINER_EXTRA_PARAMS` variable is not exported when running `gnu parallel`. 

We need that variable because it passes `--rgw-lc-debug-interval` when life cycle tests are enabled (they are enabled by default)

It also adds the `LIFE_CYCLE_INTERNAL_PARAM` variable as that one is needed when not running s3tests with the container option.

Fixes: https://github.com/aquarist-labs/s3gw/issues/474

### DIFF
--- a/tools/tests/s3tests-runner.sh
+++ b/tools/tests/s3tests-runner.sh
@@ -279,6 +279,8 @@ _main() {
     export TMPFILE
     export OUTPUT_DIR
     export PARALLEL_HOME
+    export CONTAINER_EXTRA_PARAMS
+    export LIFE_CYCLE_INTERVAL_PARAM
 
     mkdir -p "$PARALLEL_HOME"
     parallel --record-env
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -16,7 +16,13 @@ jobs:
         uses: actions/checkout@v3
         with:
           path: s3gw
-          submodules: true
+          submodules: false
+
+      - name: Checkout ceph
+        uses: actions/checkout@v3
+        with:
+          repository: aquarist-labs/ceph
+          path: s3gw/ceph
 
       - name: Checkout s3tests
         uses: actions/checkout@v3


tools/tests/s3tests-runner.sh
```diff
@@ -279,6 +279,7 @@ _main() {
     export TMPFILE
     export OUTPUT_DIR
     export PARALLEL_HOME
+    export CONTAINER_EXTRA_PARAMS
```
m-ildefons, line 282: You probably need to export `LIFE_CYCLE_INTERNAL_PARAM` too, to make it work when it's not running the s3gw in a container.

m-ildefons (changes requested):
Was `LIFE_CYCLE_INTERNAL_PARAM` left out on purpose?

tools/tests/s3tests-runner.sh
0xavi0: Yes, although CI seems to be always running with the container option we need that variable also.
Added

m-ildefons: Yes. CI always uses containers, but locally you might not want to build a container before running s3tests. In that case you can just point the script at your ceph source tree and if it finds the radosgw binary, it will use that.
With the parallelization, using this script over just executing the s3tests directly would give a speed advantage.

m-ildefons (dismissed):
LGTM

m-ildefons (approved):
LGTM, although I'm not too sure about checking out the ceph repo in the directory where the submodule should live, but that shouldn't give us too much trouble for now.

PR #497, opened 2023-05-05
SFS ADRs based on Design Review Session 2023-05-05

irq0:
Signed-off-by: Marcel Lauhoff <marcel.lauhoff@suse.com>

### DIFF
--- a/docs/decisions/0010-sfs-versioning.md
+++ b/docs/decisions/0010-sfs-versioning.md
@@ -1,0 +1,142 @@
+# SFS Versioning
+
+Refines ADR 0003-SFS on the object state machine and database columns.
+
+## Database
+
+In this document we only look at tables Objects and Versioned Objects.
+
+An *object* has a name, id, and reference to a *bucket*
+
+A *object version* has an id, checksum, {create, commit, delete}_time,
+mtime, size, **state**, **type**, ETag, serialized attributes, etc.
+
+An *object* is a group of versions identified by bucket and name.
+
+An *object* has one or more *object versions*. Regardless of the bucket versioning setting.
+
+Object **state** is an enum of Open, Committed, Deleted. See Object State Machine.
+
+Object **type** is an enum of Regular, Delete Marker.
+
+## Object State Machine
+
+```mermaid
+stateDiagram-v2
+    [*] --> Open : Create new version
+    Open --> Committed : Success
+    Open --> Open : Writing
+    Committed --> Deleted : Delete
+    Open --> [*] : Failure
+    Deleted --> [*] : Garbage Collection
+```
+
+*Open* - Initial state. Data is in flight. Data on disk is dirty. A
+version may stay in this state if, for example, a client fails during
+a PUT operation. A to be defined GC process cleans this up after a
+time.
+
+*Committed* - Upload finished. Data persisted. GETs and LIST
+will return this object.
+
+*Deleted* - Soft deletion state, but in terms of S3 permanent deletion.
+
+*terminal state* - the row no longer exists in the database.
+
+An object will never move back from Deleted to Committed. Deleting a
+version is treated as permanent, even though it is only permanent after
+the GC made it so.
+
+State changes also change timestamps:
+
+## Timestamps
+
+(See ADR 11 for information on the data type we store)
+
+Versioned objects store the following timestamps:
+
+*commit_time* - When the object changes to *Committed* state.
+
+*delete_time* - When the object changes to *Deleted* state.
+
+*create_time* - Set when the row is first created in *Open* state.
+
+*mtime* - A modification time passed to SFS. We follow the RADOS
+  logic: Passed a `set_mtime` and `mtime`, we persist `set_mtime`. If
+  `is_zero(set_mtime)` we take `real_clock::now()`. Return the `mtime`
+  we persisted.
+
+## Object Version Types
+
+An object version may be of special type *delete marker*, representing
+an S3 delete marker in versioned buckets.
+
+## Garbage Collection
+
+Input: object versions in deleted state.
+
+Either delete single versions or delete whole objects.
+
+In case whole objects, delete the object row if all versions are deleted.
+In a transaction delete all versions, then the object.
+
+Should a concurrent transaction add a new version either will fail on
+the foreign key constraint between version and object. In the GC we
+continue with the next object; in the create path we retry or let the
+client retry.
+
+## Operations
+
+### Deletion
+
+Deleting a version is not to be confused with hiding objects by adding
+deletion markers. Deleting an object sets the state to DELETED.
+
+### Hiding / Deletion Marker
+
+A delete to a versioned bucket without a version creates a delete
+marker. This hides the object and all its version from unversioned GETs
+and LISTS.
+
+SFS implements this with object versions having type delete marker. If
+one exists on an object, it is excluded.
+
+Deletion markers are object versions and follow the same state machine
+and garbage collection as regular versions do. They however don't have
+a size and data on the filesystem.
+
+### Create / Update
+
+(versioned, unversioned buckets) In a transaction:
+
+1. Find or create an object row (id, name, bucket)
+2. Create an object version in state OPEN with creation time of now.
+    If versioned, use the version id provided. Otherwise generate one.
+
+Receive and write data.
+
+On completion:
+(versioned buckets) set state COMMITTED of the previously created version. Record
+checksum, size, etc.
+
+(unversioned buckets) set state COMMITTED of the previously created version
+AND set all other versions DELETED.
+
+### Access: Listing, GETs
+
+Many operations expect a head, null, last or "canonical" version of
+an object. We define this as the *last* committed version, having the *latest*
+*commit_time* timestamp.
+
+If the timestamp resolution is too low to distinguish versions the highest id wins.
+
+(unversioned) Since we rely on versions to implement updates, more
+than one committed version may exist. We make this less likely by the commit rule in
+Create / Update above. Should multiple exists, the latest is the one we use.
+
+### Out of scope (for now): Versioning Suspended Buckets
+
+For now SFS won't support versioning suspended buckets. Object
+versions created while the bucket was unversioned or in versioning
+suspended state have a 'null' version id. This is not directly
+supported by this design and requires refinement in the future.
--- a/docs/decisions/0009-sfs-collection-of-high-level-design-decisions.md
+++ b/docs/decisions/0009-sfs-collection-of-high-level-design-decisions.md
@@ -1,0 +1,23 @@
+# Collection of High Level Design Decisions
+
+Use soft deletion. Mark versions deleted, let a garbage collector hard
+delete *later*.
+
+Non-versioned objects are a special case of versioned objects. They
+generally follow the same logic.
+
+The SFS SQLite database is the source of truth. Example: If we delete
+an object, we first modify the database then the filesystem. Example:
+Serve metadata (object size, mtime, ..) from the database rather
+than stat() the file.
+
+SQLite transactions are atomic. Filesystem operations maybe. Both combined are not.
+Orphaned files on the filesystem are acceptable and countered by an offline fsck tool.
+
+Use Ceph bufferlist encodings of data structures in the database where
+we don't have to query individual fields. Example: object attrs is
+bufferlist encoded, deletion time is not. With this leverage Ceph data
+structure versioning support as much as possible.
+
+Use negative return value style error handling not exceptions. This
+follows from the Google C++ style guide and the general RGW style.
--- a/docs/decisions/0011-sfs-timestamps.md
+++ b/docs/decisions/0011-sfs-timestamps.md
@@ -1,0 +1,62 @@
+<!-- #cSpell:words multicol datefunc -->
+# SFS Timestamps
+
+## Context and Problem Statement
+
+We need to create, handle and persist timestamps.
+
+In the RGW space we have ceph_time.h and std::chrono.
+
+SQLite represents time as ISO-8601, Julian day or unix timestamps.
+Stored in TEXT, REAL or INTEGER data types. [SQLite doc:
+data types](https://www.sqlite.org/datatype3.html). It has functions to
+work with these data types. [SQLite doc:
+datefunc](https://www.sqlite.org/lang_datefunc.html).
+
+This doc is about the conversion between RGW/sfs and SQLite space.
+
+Summary of discussion in weeks 19, 20 2023. [GH
+Comments](https://github.com/aquarist-labs/s3gw/pull/497)
+
+## Ceph time
+
+We use `ceph::real_time` as timestamps.
+
+`ceph::real_time` is a `uint64_t` nanosecond count since epoch.
+
+## How to store time in SQLite
+
+### Requirements
+
+Minimum microsecond resolution.
+
+We can leverage SQLIte range queries, sorting, indices. Ideally we can
+leverage date / time functions with minor conversion.
+
+### Considered Options
+
+Options, that don't fit our requirements:
+
+- *ISO8601* strings. Not enough resolution.
+- *UNIX time* - Not enough resolution
+
+- *multicol* - Store seconds as *ISO8601* or *UNIX time* (SQLite
+  functions work directly). Store nanoseconds in a second column.
+  Queries awkward.
+- *int64 ns* - Store as int64, nanosecond resolution, convert from
+  uint64. Max value up to year 2262. Queries work.
+- *int64 us* - Store as int64, microsecond resolution, convert from
+  uint64. Max value up to year 2554. Queries work.
+- *hex* - Store as 16 char hex string. Full `ceph::real_time` range.
+  Conversion cost. Queries work.
+- *uint64* - Squeeze `uint64_t` into SQLite `INTEGER` type
+  representation. No queries.
+- *blob* - Store `ceph::real_time` as an SQLite blob. Can't query and
+  index as easily as *hex* or *int64* options.
+
+### Decision Outcome
+
+We choose *int64 ns*, because it meets all criteria.
+
+It requires minimal conversion. We can live with the year 2262
+limitation.
--- a/docs/dicts/s3gw.dict
+++ b/docs/dicts/s3gw.dict
@@ -53,3 +53,4 @@ pytest
 proxying
 COSI
 objectstorage
+bufferlist
--- a/docs/dicts/s3gw.dict
+++ b/docs/dicts/s3gw.dict
@@ -54,3 +54,4 @@ proxying
 COSI
 objectstorage
 bufferlist
+ETag
--- a/docs/dicts/s3gw.dict
+++ b/docs/dicts/s3gw.dict
@@ -55,3 +55,4 @@ COSI
 objectstorage
 bufferlist
 ETag
+chrono


giubacc:
I've found what the `null` version is: https://docs.aws.amazon.com/AmazonS3/latest/userguide/AddingObjectstoVersionSuspendedBuckets.html

Basically the null version occurs whenever a versioned bucket becomes unversioned because versioning has been suspended.
In this case, if you PUT an object, its version becomes the null version, that is, a fake id denoting the null version (that can be overwritten by subsequent PUTs).
If you delete an object on a versioned bucket a it is needed to check if the null version exists and in case it will be deleted.
So the null version is the transposition of the "regular" object of an unversioned bucket in the world of the versioned buckets.
I think it could be worth to add this information to the ADR.

docs/decisions/0010-sfs-versioning.md
jecluis, line 1: maybe move this to `docs/design` ? Makes more sense I think.

jecluis (changes requested):
LGTM, but I'd like the doc to be moved to `docs/design/` instead. That, or we'll need to bring the other doc there here. I do think there's an advantage in keeping design docs all in the same place though.

irq0:
> LGTM, but I'd like the doc to be moved to `docs/design/` instead. That, or we'll need to bring the other doc there here. I do think there's an advantage in keeping design docs all in the same place though.

I don't have a strong opinion on either one directory or two. 
Only thing I would love to have is a README that tells us what goes where. I can't intuitively distinguish between a "architecture decision record" and a design document.

So, what goes where? 

**doc/decisions**

- how we work (0000-use-markdown-any-decision-records.md, 0001-reorganize-documentation.md, 0004-release-captain.md, 0005-project-labels.md, 0006-git-branches-tags.md, 0007-release-process.md, 0008-pull-requests-code-review-guide.md)

**doc/design**
- how we want the software to work (0002-s3gw-storage-backend.md, 0003-sfs.md, 0009-sfs-collection-of-high-level-design-decisions.md, 0010-sfs-versioning.md)


I think doc/decisions is too confusing. Suggest the following:

- doc/decisions/design - how we want the software to work
- doc/decisions/how-we-work - how we work docs from above

jecluis:
>     * how we work (0000-use-markdown-any-decision-records.md, 0001-reorganize-documentation.md, 0004-release-captain.md, 0005-project-labels.md, 0006-git-branches-tags.md, 0007-release-process.md, 0008-pull-requests-code-review-guide.md)
> 
> 
> **doc/design**
> 
>     * how we want the software to work (0002-s3gw-storage-backend.md, 0003-sfs.md, 0009-sfs-collection-of-high-level-design-decisions.md, 0010-sfs-versioning.md)
> 
> 
> I think doc/decisions is too confusing. Suggest the following:
> 
>     * doc/decisions/design - how we want the software to work
> 
>     * doc/decisions/how-we-work - how we work docs from above

That division sounds good to me. We might want to renumber them though, per subdir? @jhmarina thoughts?

jecluis:
@irq0 in the meantime, given that's a tangential topic, lets just go ahead and merge this. That can be handled on a different PR.

jecluis (dismissed):

jecluis:
> @irq0 in the meantime, given that's a tangential topic, lets just go ahead and merge this. That can be handled on a different PR.

And I mean "merge once everybody has reviewed and is happy (or not too sad)" :)

docs/decisions/0010-sfs-versioning.md
```diff
@@ -0,0 +1,114 @@
...
+checksum, size, etc.
+
+(unversioned buckets) set state COMMITTED of the previously created version
+AND set all other versions DELETED.
+
+### Access: Listing, GETs
+
+Many operations expect a head, null, last or "canonical" version of
+an object. We define this as the *last* version, having the *latest*
+creation time.
```
0xavi0: The current implementation uses de `id` of the version. The greater `id` is the latest version. (`id` is the primary key of the versioned object table)
If we want to implement this based on time... Wouldn't it be better to use modification time instead to creation time?. In case that 2 threads are putting a version in parallel, using the creation time means the last trying to write wins, while using the modification time means the *last* that finished writing wins. I think it makes more sense that the one winning is the last writing?

Also, if we want to query what's the latest version based on creation/modification time we might need to change the schema, because right now those time columns are stored as `ceph::real_time` which is stored as a `BLOB` in the database. 

Problem is that _maybe_ the `SQLite` data structures have not enough precision?


**From the SQLite documentation**
```
SQLite does not have a storage class set aside for storing dates and/or times. Instead, the built-in [Date And Time Functions](https://www.sqlite.org/lang_datefunc.html) of SQLite are capable of storing dates and times as TEXT, REAL, or INTEGER values:

TEXT as ISO8601 strings ("YYYY-MM-DD HH:MM:SS.SSS").
REAL as Julian day numbers, the number of days since noon in Greenwich on November 24, 4714 B.C. according to the proleptic Gregorian calendar.
INTEGER as Unix Time, the number of seconds since 1970-01-01 00:00:00 UTC.
```

jecluis: > Also, if we want to query what's the latest version based on creation/modification time we might need to change the schema, because right now those time columns are stored as `ceph::real_time` which is stored as a `BLOB` in the database.

On the multipart work I'm using timestamps to store a `ceph::real_time` instead, which I think is stored as a double. The conversion is trivial to do, and we get to use those fields. I didn't think about the necessary precision though, I should double check that.

irq0: Looks like we need to standardize on a time format and add that to the high level doc.

For reference: 
`ceph::real_time` is `std::chrono::time_point<real_clock>`
`real_clock` is Ceph's high resolution real time clock. Nanosecond resolution.

I'd love to keep as much resolution as we can. Minimum microseconds. At the same time leveraging sqlite indices, range queries, sorting is important too.

Options: 
1. Store as ISO8601 strings. We loose resolution. Values are monotonically increasing. order by and range queries work nicely. We can use sqlite datetime functions.
2. Store as integer. Unix time is second resolution = meh. So store micos? nanos? SQLite has max 64bit integers. I don't thing struct timespec fits good into this.
3. Store as multiple columns. ISO8601 string + nanoseconds part or two integers. First for search, second for resolution. 
4. Store as blob. I read somewhere[42] that sqlite does comparisons based on memcmp on this. If we encode the time data right, we can have it comparable and monotonically increasing. No idea if we can do BETWEEN queries and how it indexes. Probably the least desirable format. 

What else?


[42] https://www.sqlitetutorial.net/sqlite-data-types/

0xavi0: IMO, we should try to keep the same precision that `rgw` has... (nanos). 
I would prefer not to have things in more than 1 column.

Isn't 2 ^ 64 enough? .... it should cover up to year 2554, right? 
I think SQLite stores signed INTEGERS, though....

jecluis: I'm using `ceph::real_clock::to_double(mp.mtime)` to obtain a double, which I assumed would capture a high resolution clock value, but now I'm not so sure. It seems to translate to

`std::chrono::duration<double>(t.time_since_epoch()).count();`

which seems to be using the default for `std::chrono::duration`, that being "seconds" as far as I can tell.

jecluis: then again, we _could_ rely on `std::chrono` directly when converting to/from the database, and apply whatever precision we want.

0xavi0: `ceph::real_time` is, if my understanding is correct, represented by a `uint64_t`. 

`ceph::real_time` time_point is defined as: 
```c++
typedef std::chrono::time_point<real_clock> time_point;
```

`ceph::real_clock` is:
```c++
class real_clock {
public:
  typedef timespan duration;
  typedef duration::rep rep;
  typedef duration::period period;
```
and `timespan` is: 
```c++
typedef std::chrono::duration<ceph::rep, std::nano> ceph::timespan
```

and... finally: `ceph:rep`: 
```c++
namespace ceph {
// Currently we use a 64-bit count of nanoseconds.

// We could, if we wished, use a struct holding a uint64_t count
// of seconds and a uint32_t count of nanoseconds.

// At least this way we can change it to something else if we
// want.
typedef uint64_t rep;
```

So the actual `ceph::real_clock` doesn't have the resolution that we can get from `struct timespec` (16 bytes), it fits in 64 bits by design. (I confirmed it printing out sizeof(now), being now = ceph::real_clock::now() )

0xavi0: Actually... there is a sqlite_uint64.

https://www.sqlite.org/c3ref/int64.html

and it is also present in `sqliteorm`:
```c++
namespace sqlite_orm {

    using int64 = sqlite_int64;
    using uint64 = sqlite_uint64;
```

So...taking into account that `ceph::real_clock` is represented as a uint64_t,  storing a `ceph::real_clock::time_point` value would be something like:
```c++
sqlite_orm::uint64 db_value = std::chrono::duration_cast<std::chrono::nanoseconds>(time_value.time_since_epoch()).count();
```

and reading a `ceph::real_clock::time_point` from the database:
```c++
auto rgw_value = ceph::real_clock::time_point(std::chrono::nanoseconds(value_from_db));
```

0xavi0: meh. I suspected `sqlite` worked internally with signed 64 bits integers and I did a small test.

I stored 5 objects, with the following values in a `test` field:
* 0
* std::numeric_limits<int64_t>::max() - 1 
* std::numeric_limits<uint64_t>::max() - (std::numeric_limits<uint64_t>::max()/4)
* std::numeric_limits<uint64_t>::max()
* 1234555555

I did the following query:
```c++
auto objs = storage.get_all<DBObjs>(
        sqlite_orm::order_by(&DBObjs::test).desc()
    );
```
and the result is:
```
9223372036854775806
1234555555
0
18446744073709551615
13835058055282163712
```

so sign is taken into account.
I declared the `test` field as `sqlite_orm::uint64`

I had no issue storing and retrieving uint64 values... but the problem is when we try to get things in order... or mix/max values.

I also tried to do 
```c++
storage.max(&DBObjs::test);
```
and it returns the maximum signed 64 bits integer: 9223372036854775806

I'll investigate a bit more... with signed 64 bits numbers we can maybe work up to 63 bits which would give us time values from epoch until year 2262 (if I'm not mistaken), maybe the `s3gw` is still in use by that year... who knows... :)

0xavi0: I've been testing another possible solution. I prefer the uint64 storage, but as observed above we can't rely on it. (and I suspect we can loose precision storing a `double`)

So... for storing the timestamp... I tested to store the hex string representation of the uint64_t that represents the `ceph::real_clock::time_point` (max  uint64 would be stored as `ffffffffffffffff`, 0 would be represented as `0000000000000000`.
That means... fixed length strings  that `sqlite` is able to sort and compare and so on.

Obviously, in terms of latency, this option is slower because there is a layer of conversion, but strings will be always 16 char strings. 

For testing it I used this for the uint64_t -> string conversion:
```c++
auto db_value = fmt::format("{:0>16x}", uint64_value);
```

I saw fmt::scan is still in the test folder of libfmt so I simply used `std::stoull` for the str to uint64_t conversion.
```c++
auto rgw_value = std::stoull(str_db_value, nullptr, 16);
```

This time the test was correct calling the same `order_by` query:
```
18446744073709551615
13835058055282163712
9223372036854775806
1234555555
0
```

and `storage.max()` returns 18446744073709551615, which is correct.

0xavi0: And finally... something I wanted to investigate long ago when all this started and it's custom type binding in `sqlite_orm`


The idea is to be able to do this:
```c++
struct DBObj {
  uint id;
  ceph::real_clock::time_point test_time;
};
```

and work with `ceph::real_clock::time_point` custom type in `sqlite` as if it was an internal type.

This is the code for specialising `sqlite_orm` to support `ceph::real_clock::time_point` (using the 16 chars long hex string conversion from the example above):
```c++

namespace ceph {
static std::string to_uint64_encoded(const ceph::real_clock::time_point& t) {
    uint64_t nanos = std::chrono::duration_cast<std::chrono::nanoseconds>(t.time_since_epoch()).count();
    return fmt::format("{:0>16x}", nanos);
}

static std::optional<ceph::real_clock::time_point> from_uint64_encoded(const std::string& s) {
    std::optional<ceph::real_clock::time_point> ret;
    try {
      uint64_t nanoseconds = std::stoull(s, nullptr, 16);
      ceph::real_clock::time_point ret_time = ceph::real_clock::time_point(std::chrono::nanoseconds(nanoseconds));
      ret = ret_time;
    } catch (const std::exception & ex) {
      // Do something here, I'm just a test
    }
    return ret;
}
}

namespace sqlite_orm
{
  template<>
  struct type_printer<ceph::real_clock::time_point> : public text_printer {};

  template<>
  struct statement_binder<ceph::real_clock::time_point> {

      int bind(sqlite3_stmt* stmt, int index, const ceph::real_clock::time_point& value) const {
          return statement_binder<std::string>().bind(stmt, index, to_uint64_encoded(value));
      }
  };

  template<>
  struct field_printer<ceph::real_clock::time_point> {
      std::string operator()(const ceph::real_clock::time_point& t) const {
          return ceph::to_uint64_encoded(t);
      }
  };

  template<>
  struct row_extractor<ceph::real_clock::time_point> {
      ceph::real_clock::time_point extract(const char* row_value) const {
          if(row_value) {
              auto sd = ceph::from_uint64_encoded(row_value);
              if(sd) {
                  return sd.value();
              } else {
                  throw std::runtime_error("incorrect encoded timestamp string (" + std::string(row_value) + ")");
              }
          } else {
              // ! row_value
              throw std::runtime_error("incorrect encoded timestamp string (nullptr)");
          }
      }

      ceph::real_clock::time_point extract(sqlite3_stmt* stmt, int columnIndex) const {
          auto str = sqlite3_column_text(stmt, columnIndex);
          return this->extract((const char*)str);
      }
      ceph::real_clock::time_point extract(sqlite3_value* row_value) const {
          auto characters = (const char*)(sqlite3_value_text(row_value));
          return extract(characters);
      }
  };
}
```

Now you can do something like:
```c++

struct DBObj {
  uint id;
  ceph::real_clock::time_point test_time;
};

DBObj obj;
// get the ceph::real_clock::time_point type and work with it
// no need to do conversions in userspace code (they are done by sqliteorm specialisation)
obj.test_time = ceph::real_clock::now();
auto storage = conn->get_storage();
storage.insert(obj);

// get objs by ceph::real_clock::time_point in descending order
auto objs = storage.get_all<DBObj>(
        sqlite_orm::order_by(&DBObj::test_time).desc()
    );

// get max time
auto max_time = storage.max(&DBObj::test_time);

// get max time for id == 1999 
auto max_time_1999 = storage.max(&DBObj::test_time, 
  where(c(&DBObj::id) = 1999));
  
// select where time == whatever
ceph::real_clock::time_val some_time_we_got;
auto objs_matching_time = storage.get_all<DBObj>(where(c(&DBObj::test_time) = some_time_we_got));
```

irq0:
```c++
#include <limits>
#include <cstdint>
#include <iostream>
#include <chrono>
#include <fmt/core.h>
#include <format>


int main() {
    auto tp_max_unsigned = std::chrono::time_point<std::chrono::system_clock, std::chrono::duration<uint64_t, std::nano>>::max();
…    if (ns_since_epoch > std::numeric_limits<int64_t>::max()) { /* fail */ }   
    const int64_t ns_since_epoch_signed = static_cast<int64_t>(tp.time_since_epoch().count());

    std::cout << std::format("now: {:%Y-%m-%d %H:%M:%S}", tp) << std::endl;
    std::cout << std::format("int: {} {}", ns_since_epoch, ns_since_epoch_signed) << std::endl;

    const auto tp_again = std::chrono::time_point<std::chrono::system_clock, std::chrono::duration<uint64_t, std::nano>>(
        std::chrono::duration<uint64_t, std::nano>(ns_since_epoch_signed));
    std::cout << std::format("now again: {:%Y-%m-%d %H:%M:%S}", tp_again) << std::endl;
}
```

-> 

```
uint64_t max tp: 2554-07-21 23:34:33.709551615
int64_t max tp: 2262-04-11 23:47:16.854775807
now: 2023-05-12 17:12:03.537852957
int: 1683911523537852957 1683911523537852957
now again: 2023-05-12 17:12:03.537852957
```

[godbolt](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1DIApACYAQuYukl9ZATwDKjdAGFUtAK4sGIM6SuADJ4DJgAcj4ARpjEIADMGqQADqgKhE4MHt6%2BeqnpjgIhYZEsMXFctpj2hQxCBEzEBNk%2BfgF2mA6Z9Y0ExRHRsQlJCg1NLbmVo739peUSAJS2qF7EyOwc5vGhyN5YANQm8W60eCyECkfYJhoAgls7e5iHx8ij%2BIJXN/dm2wy7XgORzcTlGxEwrC%2Bdwe/yeLzcyAQxAEqChPz%2BAKBxyoLAIAHo0OCAHQINEwzHPYH8YgsJgEMnQu6hAj7WmhCALQ4Adisd32/P2TC8RH2BGSAH1aapxV4GOlgGF0C8ACL7d4gFBIlEaxxscWpZnA9Wa5EMVAahQAT1GmBY4t2qGQAGtSGqCOgNYjTeaQOhVnTMsCvMyAGyScUEV3G5hmr7xbAaqUco6824CwXC1CiiVS8XyxUqt0ek3akC6zD61CG47Gr2lq02u0O51R92erVmjV%2B4gBgTA0PhyNFjUx1Hxq6JpiqZPxVPp2srFnA4GHPnztsgam0ggQcxmYOCMMR1lT7MgbkWEDmACsAE0ALQ3liPszXpU3gASV9fAFlv9ehBMLllT3V0xUlKcZTlPAFUwdBOWXY5hxAVxaBTb512LNBhXhFcMMwjUtzpXczDMAdjylM8L3/B8nxfN9DlfL8bz/G9AOA0Dswg6V8zghDjhXY1UPQxk0wFNA5RZWsOx9ctK2rNxpO9C1rQIW17VoR0XWQutO19f1aiDcih2jQwx2ueNsxVfD035JT61U9TmydEdUAAdxnVMbP2CTRn2A8CCPFk5TzHYK0wVJEULMUiTk9J/nCyKEA5IlsMETzvLwKh9ggEL4vWcUIsdBAXmwZCGB8WI8GQcVTnOAhLmOYyJxAJMFgQnl9jxAAqfYqCYOh9m6vFuVVflvN8lljP2PKwsKpLQtg994lVaZHBq0RRn7Q9ByuCAYriuaisRFK0p3dqRPuNc7I3bClwEpDjSInc9zNNzzyAy8b1o19nxvd8mP/VjX3YkDSLA5J%2BLcQSN2E2dvIXHDELcZDnpIsjBA%2BnkgNVT6cc42aEvm4rXUJgrjoQRbFShmHizhrzrp8gQ/KFEVwKYYABoYQt7L0uSDU%2BGtbpklTGw0rTWywkX9J7QzjgCoLJZHMy42wXdGYIks9O7XsGCMnaIyVkBRz2snEuKqm%2BIWS7Nbu3DHo3NHXvcwVOdCLGvtfH7rz%2B18AevZjf3/UHOPZt2GBph26YYdA0Ph6FgI4JZaE4a9eD8DgtFIc0OGhyxrDVFY1kpX4eFIAhNCTpYXLDIkAE4AA4NDrrgQw0RJr3iSR4hDfROEkdPK%2BzzheAUEAkgrzOk9IOBYBgRAUFQFhkjoWJyEoNBl9XuJgC4eJKiwAA3arMAANTwTA3IAeWSRhODLmhaDU4gx4gKIh6iUJGkte/eE/5hiCWivlEbQnRJ5l03mwQQV8GC0B/lPUgWBaRGHEAg/A4IuiH0wGPBBmBVCdGFBsLOzJqhD1OFEHsgCPBYCHgQYgZxf5LCoAYYAChz6XxvnfbgvB%2BCCBEGIdgUgZCCEUCodQCDdCVAMEYFA1hrD6DwFEMesBmBsBAFEVAaRIxYLiNwJu09cqsHYDibRQw6GyidO1bOyRag4PvO8I4ypTD50sGYLOHQujOAgK4CYfhKjBFCAMMoQxKj5AyAIXxeQ0jhIYLMQYFQqg1G6GMZonhWh6A8bUHoTQ4nBISdMcYaTJi2BSbk%2BYXAlgKCLusRYfcOBp1IBnLOOd9iqAbiGe8YZ9jAGQMgfYe8iRcByrgQgJBGL7wWLwSeWgrEuUkHXIkIYG51xDCGLkXIQxcA0MssMdSB6kBYCASQGgiTXlfFwKQXBrxnOvBc6QTTeA51HuPculclizwXouZIwp14QE3ivegxBwhGM4G0jpXSel9K4GYIkZheBwVGfQj0lReHCFEOIIRqLRFqCHpI0gbkezJEYXUhpDzh4cCvsKb5LJUDZTBZ0yQ3Ten9PiIMnKHgt6AvGRUqZbylgIAhFgOIHI9m8EOdeBZkguSSAbg3a8DdJBSple0xpQ8nm2BedMqupAXLxGvESDQXA67bPiCsuukgQzXktXU%2BIg8EHqq1VYlOHA4WqvtSPV5U8rE6PSM4SQQA)


We should be fine with conversion to int64 keeping nanosecond resolution, not?

irq0: > And finally... something I wanted to investigate long ago when all this started and it's custom type binding in `sqlite_orm`

Very cool! Wanted to try the binding support for a while as well :). Looks good. It could replace the conversion layer we have if we add bindings for the ceph types, right?

0xavi0: This would move the conversion to another place, but the conversion code will be still needed.
We won't see any performance gain. (and the code to be maintained will be highly templated code, which not everybody loves)

This is something I wanted to investigate since the very beginning but given that we didn't had the time I could not go deeper. 

I think it makes user code cleaner, because we work with our own types.

Anyway, I would maybe do it for the timestamps in this iteration and will try to change the rest when the versioning stuff is done. I don't think this is as urgent as versioning work.

0xavi0: Test code for storing `ceph::real_clock::time_point` as `int64` and checking that values stored or read are not negative (throws a `std::runtime_error` in that case) 

I can `order_by` and use `max` and `min` from the `sqlite` storage class same as with the 16 chars hex string representation.

Trying to pass a `time_point` like `ceph::real_clock::time_point(std::chrono::nanoseconds(std::numeric_limits<uint64_t>::max()))`  won't be accepted. 
```
Error converting ceph::real_clock::time_point to int64. Nanoseconds value: 18446744073709551615 is out of range
```

```c++
namespace ceph {
static int64_t to_int64(const ceph::real_clock::time_point& t) {
    uint64_t nanos = std::chrono::duration_cast<std::chrono::nanoseconds>(t.time_since_epoch()).count();
    // we check that the value is not greater than int64 max
    if (nanos > std::numeric_limits<int64_t>::max()) {
      std::stringstream oss;
      oss << "Error converting ceph::real_clock::time_point to int64. Nanoseconds value: " << nanos << " is out of range";
      throw std::runtime_error(oss.str());
    }
    return static_cast<int64_t>(nanos);
}

static std::optional<ceph::real_clock::time_point> from_int64(int64_t int_time) {
    std::optional<ceph::real_clock::time_point> ret;
    if (int_time < 0) {
      // to ensure that we stick to the int64 positive range.
      std::stringstream oss;
      oss << "Error converting int64 nanoseconds value to ceph::real_cock::time_point. Value: " << int_time << " is out of range";
      throw std::runtime_error(oss.str());
    }
    uint64_t uint64_nanos = static_cast<uint64_t>(int_time);
    ceph::real_clock::time_point ret_time = ceph::real_clock::time_point(std::chrono::nanoseconds(uint64_nanos));
    ret = ret_time;
    return ret;
}
} // namespace ceph

namespace sqlite_orm
{
  template<>
  struct type_printer<ceph::real_clock::time_point> : public integer_printer {};

  template<>
  struct statement_binder<ceph::real_clock::time_point> {

      int bind(sqlite3_stmt* stmt, int index, const ceph::real_clock::time_point& value) const {
          return statement_binder<uint64_t>().bind(stmt, index, to_int64(value));
      }
  };

  template<>
  struct field_printer<ceph::real_clock::time_point> {
      std::string operator()(const ceph::real_clock::time_point& t) const {
          auto int_value = ceph::to_int64(t);
          return std::to_string(int_value);
      }
  };

  template<>
  struct row_extractor<ceph::real_clock::time_point> {
      ceph::real_clock::time_point extract(int64_t row_value) const {
          if(row_value < 0) {
             throw std::runtime_error("incorrect encoded timestamp int64. (" + std::to_string(row_value) + ") is out of range");
          }
          auto sd = ceph::from_int64(row_value);
          if(sd) {
            return sd.value();
          } else {
            throw std::runtime_error("incorrect encoded timestamp int64 (" + std::to_string(row_value) + ")");
          }
      }

      ceph::real_clock::time_point extract(sqlite3_stmt* stmt, int columnIndex) const {
        std::cout << "Extracting value" << std::endl;
        auto int_value = sqlite3_column_int64(stmt, columnIndex);
        return this->extract(int_value);
      }
  };
}

Comment (0xavi0) [reply docs/decisions/0010-sfs-versioning.md]: As I know there is a lot of noise I'm making a summary:

Options I see so far.
* Store timestamps as `int64` (which will limit our timestamps range up to year 2262)
* Store timestamps as 16 chars hex string (we can store the whole uint64 range and, if in the future upstream adds more precision we can increase the number of chars... This comes with a conversion code that adds latency)
* Store as `uint64` (but we cannot query by timestamps because sqlite takes sign into account for order_by and max, main)
* Keep storing it as a `blob` if we don't need to query by timestamp. 


And... for deciding how to identify what's the last version of an object.
* Max creation time
* Max modification time (or another field that we create so the last thread writing wins)
* Max version id (I think we can go for this in favor of max creation time, because they are equivalents and we don't need to change the timestamps code. When we set creation time we also create a new version id, so the greater one will still be the latest one)

Comment (irq0) [reply docs/decisions/0010-sfs-versioning.md]: I'd like to add option int64, but with microsecond resolution (max 2554-07-21 23:34:33.709551).

**Comment by irq0** [comment]
Last push adds: 
Timestamp discussion as adr 11
ADR 10: Timestamps to record, refinement of the last version rule.

File: docs/decisions/0010-sfs-versioning.md [hunk 2]
```diff
@@ -0,0 +1,139 @@
...
+
+*commit_time* - When the object changes to *Committed* state.
+
+*delete_time* - When the object changes to *Deleted* state.
+
+*create_time* - Set when the row is first created in *Open* state.
+
+Note: There is an mtime / modification time on versioned objects. It
+is unrelated to the object state machine. It's value is "owned" by RGW
+and only stored by us.
```
0xavi0: We could add the logic implemented in `rados`
```c++
if (real_clock::is_zero(meta.set_mtime)) {
    meta.set_mtime = real_clock::now();
  }
...
modify some meta object with set_mtime
persist metadata into the bucket index with set_mtime's value
...
  if (meta.mtime) {
    *meta.mtime = meta.set_mtime;
  }
```

irq0:
Now has a bit about the set_mtime / mtime

0xavi0 (approved):
lgtm

giubacc (approved):

jecluis (approved):

PR #528, opened 2023-05-16
tools: Add s3tests runner and analysis tools

irq0:
Yet another S3 tests runner.. This one tailored to analysis (with Datasette) and written in Python. 

Datasette instance for latest s3gw results: https://s3gw-s3test-results.fly.dev/results

## Components

s3tests runner: Run tests isolated against fresh containers. Collect test and container logs. Outputs a JSON file with logs and pytest metadata.

to_sqlite: Convert runner JSON to a SQLite database suitable to use with Datasette

metadata.yml: Datasette metadata config with canned queries and predefined facets.

Let pre-commit cover the new Python code with the existing releasetool checks.

### DIFF
--- a/tools/s3tests/runner.py
+++ b/tools/s3tests/runner.py
@@ -1,0 +1,527 @@
+#!/usr/bin/env python3
+"""
+A s3-tests runner tailored to s3gw containers with parallel execution,
+result and log gathering
+"""
+
+import io
+import json
+import logging
+import multiprocessing
+import os
+import pathlib
+import random
+import subprocess
+import sys
+import tarfile
+import tempfile
+import time
+from contextlib import suppress
+
+import click
+import docker
+import radosgw
+import requests
+
+LOG = logging.getLogger("s3tr")
+
+# How long an individual pytest may run in seconds
+PYTEST_TIMEOUT_SEC = 100
+
+# From vstart.sh::do_rgw_create_users
+S3TESTS_USERS = {
+    "s3 main": {
+        "uid": "0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef",
+        "access_key": "ABCDEFGHIJKLMNOPQRST",
+        "secret_key": "abcdefghijklmnopqrstuvwxyzabcdefghijklmn",
+        "display_name": "youruseridhere",
+        "email": "s3@example.com",
+        "user_caps": "user-policy=*",
+    },
+    "s3 alt": {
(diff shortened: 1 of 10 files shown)


tools/s3tests/to_sqlite.py
```diff
@@ -0,0 +1,220 @@
...
+    """
+    pass
+
+
+@click.group()
+def datasette():
+    """
+    Open results in datasette
+    """
+    pass
```
m-ildefons, line 117: Is this TODO or leftovers?

tools/s3tests/runner.py
```diff
@@ -0,0 +1,527 @@
...
+        for retry in range(20):
+            logs = self.container.logs()
+            if logs:
+                return logs.decode("utf-8")
+            LOG.info(f"no logs for {self.container} after {retry} tries")
+            time.sleep(2 * (retry + 1))
+        LOG.warning(
+            f"no logs for {self.container} after 20 retries. returning not available"
+        )
+        return "not available"
```
m-ildefons, line 190: As I understand this method is unreliable and we rely on the `logfile` method to retrieve the logs. Why keep this around?

tools/s3tests/runner.py
```diff
@@ -0,0 +1,527 @@
...
+        "1",
+        "--err-to-stderr",
+        "1",
+        "--log-max-recent",
+        "1",
+        "--debug-rgw",
+        "10",
+        "--rgw-frontends",
+        f'"beast port={port}"',
+        "2>&1 1> /log",
```
m-ildefons, line 104: Why rely on bash redirection in the container instead of using `--log_to_file` and `--log_file` arguments for radosgw to configure it to log to the file directly itself?

m-ildefons (approved):
Looks pretty good to me. A few small questions remain, but nothing worthy of a groundbreaking discussion.

tools/s3tests/runner.py
irq0: Tried them first, did not work as I wanted. AFAIR I wanted line buffering and couldn't get it (the reason the stdbuf is there)

tools/s3tests/to_sqlite.py
irq0: Neither. See below (@to_sqlite.command() and @datasette.command()) and in add_command in s3tr.py.
Its basically a click grouping thing

tools/s3tests/runner.py
irq0: As a note to the future dev wondering about logfile() :)

PR #666, opened 2023-08-09
ci: run unittests in the nightly test suite

### LINKED ISSUE #664: [Backlog] CI: Enable Unittests for the nightlies
### What needs to be done

Run the unittests on the nightlies.

### Why it needs to be done

They have been forgotten

m-ildefons:
Run the unittests during testing of the nightly build.

Fixes: #664

### DIFF
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -91,8 +91,22 @@ jobs:
           DATE="$(date +%Y-%m-%d)"
           echo "tag=nightly-${DATE}" >> $GITHUB_OUTPUT
 
+      - name: Build Unittest Container
+        uses: docker/build-push-action@v4
+        with:
+          load: true
+          build-args: |
+            CMAKE_BUILD_TYPE=Release
+            S3GW_VERSION=${{ steps.date.outputs.tag }}
+            NPROC=16
+          tags: |
+            s3gw-unittests
+          target: s3gw-unittests
+          file: s3gw/Dockerfile
+          context: s3gw
+
       - name: Build Nightly Container
-        uses: docker/build-push-action@v3
+        uses: docker/build-push-action@v4
         with:
           push: true
           build-args: |
@@ -105,6 +119,10 @@ jobs:
             quay.io/s3gw/s3gw:nightly-latest
           file: s3gw/Dockerfile
           context: s3gw
+
+      - name: Run Unittests
+        run: |
+          docker run --rm s3gw-unittests
 
       - name: Run S3tests
         env:


.github/workflows/nightly.yaml
```diff
@@ -91,8 +91,22 @@ jobs:
           DATE="$(date +%Y-%m-%d)"
           echo "tag=nightly-${DATE}" >> $GITHUB_OUTPUT
 
+      - name: Build Unittest Container
+        uses: docker/build-push-action@v4
+        with:
+          load: true
+          build-args: |
+            CMAKE_BUILD_TYPE=Release
+            S3GW_VERSION=${{ steps.date.outputs.tag }}
```
irq0, line 100: Does this set radosgw --version? If so, can haz for release containers?

.github/workflows/nightly.yaml
```diff
@@ -106,6 +120,10 @@ jobs:
           file: s3gw/Dockerfile
           context: s3gw
 
+      - name: Run Unittests
+        run: |
```
irq0, line 124: Nit: line break unnecessary

.github/workflows/nightly.yaml
```diff
@@ -91,8 +91,22 @@ jobs:
...
 
+      - name: Build Unittest Container
+        uses: docker/build-push-action@v4
+        with:
+          load: true
+          build-args: |
+            CMAKE_BUILD_TYPE=Release
+            S3GW_VERSION=${{ steps.date.outputs.tag }}
+            NPROC=16
+          tags: |
```
irq0, line 102: Nit: line break unnecessary

irq0 (approved):
+1 if the current build fail on quay login is out of scope ;)

.github/workflows/nightly.yaml
m-ildefons: No, but this will be persisted in the container image as a label named `Version`. We're doing it for release containers as well.
The container labels will survive image re-taggings, so even if you do a
`docker tag quay.io/s3gw/s3gw:v0.19.0 irq0/foo:bar`, the image will contain a `Name` label with value `s3gw` and a `Version` label with value `v0.19.0`. They can be queried e.g. from `docker inspect`

.github/workflows/nightly.yaml
m-ildefons: The docker/build-push-action parses these things in a special way:
https://github.com/docker/build-push-action#inputs
This works and I'm not keen on fumbling around to make it work without the line break.

jecluis:
@m-ildefons this is failing - is there anything for us to do about it, or is it the usual secrets thing from an outside of org repo? If the latter, how do you want to proceed to merge this? Maybe push to an org branch instead and have it build properly?

m-ildefons:
This is the usual PR from fork problem :roll_eyes:

PR #673, opened 2023-08-17
ADR: document using POST instead of GET for some s3gw-ui backend operations

jecluis:
Additionally, move the UI backend design ADR to the decisions directory, given we are not keeping any additional design-related ADRs separate anyway.


l-mb (dismissed):
Looks good to me, thanks for the explanation. (Fairly common approach based on a quick review of other projects.)

giubacc (dismissed):
it seems fair, lgtm

docs/decisions/0014-ui-backend-api.md
```diff
@@ -0,0 +1,47 @@
...
+## Using POST instead of GET for some operations
+
+While we should strive to ensure that we are compliant with the RESTful API
+paradigm, in some instances this is not viable.
+
+One such instance is when we have a lot of information that needs to be provided
+to the API during a GET call. In such a situation it would make sense to rely on
+the GET request's body to pass the additional context we need, but unfortunately
+browsers tend to drop the body for GET requests before sending the request to
+the server.
```
votdev: Please reference to the specs here:

- https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/send
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/GET
- https://xhr.spec.whatwg.org/#the-send()-method

docs/decisions/0014-ui-backend-api.md
```diff
@@ -0,0 +1,47 @@
...
+S3 protocol allows object names to contain certain characters that are
+considered reserved for URLs, meaning we would have to URL-encode them, putting
+additional pressure on the limit we have already established. In the worst case
+scenario, an object name composed solely of reserved characters would take three
+times as many characters after being URL-encoded than its original form. That by
+itself is more than the available characters we have for a URL if the object has
+the maximum number of allowed characters of 1024.
+
+The second approach, we believe, is less obvious. Could we pass these values as
+HTTP headers? We think so. But we find that ugly and not at all obvious.
```
m-ildefons: Why is this way of "ugly and not obvious" more ugly and less obvious than the deviation from RESTful API design?

giubacc: Personally I find passing potential huge data via http headers ugly.
Better use a POST and a json payload (less ugly!)

jecluis: Yeah. What @giubacc said.

m-ildefons (approved):
lgtm

Issue #361, opened 2023-03-02
Document current HA model (Epic)


l-mb:
### What needs to be done

Document the high-availability guarantees of s3gw on top of Longhorn.

- What is the expected fail-over time in response to a node/process failure
- What does it depend on (LH setup, config options, ingress used ...)

### Why it needs to be done

Users of storage solutions expect to understand the HA model of their solutions to gauge whether it's an appropriate choice for their workloads.

This is a first step in identifying what s3gw has as-is. What's the reasonable/best thing we can do today?

Once we have that, we should have follow ups on how to improve it (based on user/PM requirements) to enable further use cases. We should, once we have this documented, also consider if we can test/validate these claims.

### Acceptance Criteria

Blog post, possibly part of the project's official documentation.

This could initially be implemented similarly to Longhorn or Rancher Telemetry. Perhaps there is an opportunity for code sharing.

### Tasks
- [ ] #685
- [ ] #733
- [ ] #803
- [ ] #841

giubacc:
Is there an idea/expectation when to allocate this activity in the context of LH integration?
The topic seems quite vast (and important).

jhmarina:
Yes, @giubacc this is planned for Milestone 4 of the Experimental Schedule, which is our v0.22.0 milestone.

jecluis:
@giubacc will be doing some initial research into what we can get out of the underlying system, and its limitations, before we evaluate what are our options to move this forward.

giubacc:
what still needs to be done:

- testing the s3gw when there is an actual node fault (not the taint but a more drastic `poweroff -f`)
- integrate a client in the restart loop and measure how many ops are failing/succeding/hanging
- finalize the research PR
- make an ADR simply pointing the research and decretating the HA model for the s3gw.

other?

jecluis:
@giubacc if there are pending tasks in this Epic (which I'm sure there are, at least with regard to documentation), please add them as tasks to this issue.

giubacc:
There is a on-going task (medik8s research) regarding the HA, I think we will have to wait until the definitive path for HA will be set, I'd keep this opened because we will have to write something in the documentation at some point, but now we still don't know exactly what

jecluis:
@giubacc please file an issue for the documentation, and whatever you think may be needed to track the effort, and add it to this epic.

PR #685, opened 2023-08-30
add High Availability research

giubacc:
High Availability research

This is a first attempt to define the direction we want to take for the HA topic with s3gw.
Feedbacks, comments, requests, considerations etc; all is good at this time. 

Related to: https://github.com/aquarist-labs/s3gw/issues/361
Signed-off-by: Giuseppe Baccini <giuseppe.baccini@suse.com>


docs/research/ha/RATIONALE.md
```diff
@@ -0,0 +1,446 @@
...
+While the Kubernetes's behavior for `EXIT-1` was the expected one, the writer
+was expecting a different behavior for the `EXIT-0` test.
+
+It is reasonable for Kubernetes to put a POD whose process exits with a code
+different than zero into `CrashLoopBackoff` state; it is less obvious that the
+same state is applied also for a POD whose process exits "normally".
+A clarification: Kubernetes puts a POD exited with `0` into `Completed` state,
+but if that POD is rescheduled, the state becomes `CrashLoopBackoff` and thus
+the behavioral schema is the same as if the process run by the POD exited
+abnormally.
```
m-ildefons: The _Pod_ is not managed on its own. It is managed through a _ReplicaSet_, which in turn is managed through a _Deployment_. A _Deployment_ is a [K8s workload primitive](https://kubernetes.io/docs/concepts/workloads/) whose _Pods_ are assumed to run indefinitely.
K8s has a workload primitive to manage _Pods_ that are not expected to run indefinitely called a _Job_. _Jobs_ may be retried (i.e. their _Pod_ will be restarted) if it exits with an error, but won't be restarted when it exits with success. You can also have a standalone _Pod_. These can run to completion as well and won't be restarted in any case.
There is also a _CronJob_, which just periodically generates a new _Job_.
So it is entirely within K8s' design that a _Pod_ managed by a _Deployment_ will be rescheduled and eventually end up in _CrashLoopBackoff_ state if it exits at all, regardless of the exit reason.

To me it looks like what you are measuring here, besides the backoff function of the scheduler, is essentially the default configuration of a K8s liveness probe. [These can be adjusted to match the workload and provide a more accurate picture of the applications state to K8s](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/) including allowances for initialization phases, temporary failure responses and the likes.

m-ildefons (commented):
Nice research. I very much like that you put some thought into what kind of failure scenarios are even within our scope and that there is a comprehensive overview of the various possible configurations with the components at hand.
Here are some comments I thought of while reading, hope you find some useful information in there

docs/research/ha/RATIONALE.md
giubacc: I agree that from the point of view of _ReplicaSet_ and _Deployment_, that are meant to keep a Pod alive, the _CrashLoopBackoff_ state when the Pod keeps exiting can make sense.

The tests implemented here are measuring the time between two events occurring on `radosgw`'s Pod: 

1. the death
2. the start  

I don't think it is similar to the concept of liveness probe (?).
The liveness probe is basically an arbitrary check to see if a Pod must be considered alive or not.

m-ildefons: The time between the death and the start is determined by how long K8s takes to a) detect the Pod is dead, which is influenced by the liveness probe and b) the time K8s takes to attempt a restart, which is influenced by the backoff function.
Additionally, the startup may be delayed if the container image has to be fetched from a registry, but that case is uninteresting for us in this discussion.
Therefore, what you are measuring here is mostly the probe and the backoff, since the actual scheduling and startup is insignificant to those two factors.
Keeping in mind that the backoff function [may also include jitter](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/wait#Backoff), you'll notice that the restarts happen close-enough at the [documented 10s, 20s, 40s, ... intervals](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#restart-policy) up until the backoff limit at five minutes is reached.

The liveness probe influences how long K8s takes to detect a dead Pod, because it can be configured to do different kinds of checks at different frequencies and tolerate different counts of failed checks before the probe considers the Pod dead.
E.g. if your liveness probe is to make a HTTP Get request against a `/health` endpoint at a frequency of once every three seconds and you tolerate up to three failed checks and you also tolerate the response to take up to 300ms, it will take almost ten seconds to detect a failed Pod.
Conversely if your probe is a simple TCP connect every second and you tolerate only one failed check and the timeout is set to 100ms, you will detect a dead Pod in well under two seconds.


The backoff function will introduce an additional wait time if a Pod fails repeatedly, increasing the wait time each time the Pod fails.
So after the first time the Pod dies, K8s will immediately restart it, but after the second time, K8s will first wait a few seconds, after the third time it will wait a few seconds more and then more and more.
This is a feature that prevents situations where mass-restarts of Pods (e.g. in the case of a node failure) lead to perpetual resource starvation. Eventually, the backoff function will cause the restart to wait long enough that a mass-restart event is distributed over time.

Lastly, there is also the startup probe. This is different from the liveness probe as it is a check dedicated to determine if the Pod has started. In cases where you have a significant startup time (e.g. databases that have to replay a write-ahead-log) this can be utilized to let the Pod run, but not consider it dead even if the liveness probe wouldn't succeed. Once the startup probe has determined that the Pod is started up, the liveness probe will be considered again.

giubacc: Let's keep in mind that these tests implemented for this research are a bit different; here we have a customized `radosgw` process that _pushes_ the `death` and `start` event to the `probe` custom tool (with the patches implemented [here](https://github.com/giubacc/ceph/tree/ha-testing)).
I designed the tests this way because I wanted to precisely measure the point in time we can say the `radosgw` is ready to accept new connections from a client.
This approach is not the classic _polling_ mode you can do with a k8s probe, but it is a _push_ (artificially reversed) way to collect death and start events.

docs/research/ha/RATIONALE.md
```diff
@@ -0,0 +1,446 @@
...
+  - [The s3gw Probe](#the-s3gw-probe)
+  - [Tested scenarios](#tested-scenarios)
+    - [EXIT-1, 10 measures](#exit-1-10-measures)
+    - [EXIT-0, 10 measures](#exit-0-10-measures)
+    - [Considerations on the tests performed](#considerations-on-the-tests-performed)
+
+We want to investigate what *High Availability* - HA - means for a project like
+the s3gw.
+
+If we identify the meaning of HA as the ability to have N independent pipelines
```
l-mb, line 33: HA is not necessarily tied to multiple independent pipelines. While High Availability (as a component of dependable computing) can be achieved through redundancy, it can also be achieved by significantly increased reliability and quality of the stack. e.g., improving the start time improves HA by lowering the Mean Time To Repair (MTTR) metric after detecting and recovering from a fault. And achieving HA in a way that *never* affects the user's operations is also not realistic, since not all faults can be masked, or the complexity required to attempt it would actually be detrimental to overall system reliability/availability (aka not cost-effective). Paradoxically, failing fast instead of hanging on an attempted fault recovery can also increase observed availability.

I'd argue our key goal is to maximize user-visible, reliable service availability, and to be tolerant with regard to certain faults (such as pod crashes, single node outages ... - the faults we consider or explicitly exclude are discussed later in the document).

docs/research/ha/RATIONALE.md
```diff
@@ -0,0 +1,446 @@
...
+this is something that could be not easy to achieve with the s3gw.
+
+With pipeline we mean all the chain from the ingress to the `PV`:
+
+```mermaid
+flowchart LR
+    ingress(ingress) ==> radosgw(radosgw) ==> PV((PV))
+```
+
+The difficulty with a project like the s3gw is due to the fact that basically
```
l-mb: Actually, this is also an advantage, due to its lower complexity. Active/active syncing of all operations is non-trivial, and an active/standby pattern is much simpler to implement with lower overhead in the absence of faults.

docs/research/ha/RATIONALE.md
```diff
@@ -0,0 +1,446 @@
...
+we have one process: the `radosgw`, that has the *exclusive* access to one resource:
+the Kubernetes `PV` where the S3 buckets and objects are persisted.
+
+Speaking in theory, 3 HA models are possible with s3gw:
+
+1. **Active-Active** pipelines
+2. **Active-Passive** pipelines
```
l-mb: I'd probably call this "warm standby" based on previous projects, to avoid confusion.

Most HA clusters consider "active/passive" fail-over to mean bringing up the resources on a fresh node while strongly guaranteeing there's only ever not more than one instance.

(A reliable HA cluster will rather risk service unavailability than risk violating that constraint, since the divergence and conflict between the data sets would lead to corruption and data loss, unless the system's got an reliable way of merging them.)

docs/research/ha/RATIONALE.md
```diff
@@ -0,0 +1,446 @@
...
+3. **One Active** pipeline (no Passive pipelines preloaded)
```
l-mb: I'd call this "active/passive failover", see above

docs/research/ha/RATIONALE.md
```diff
@@ -0,0 +1,446 @@
...
+
+## Active-Active pipelines
+
+This is the model that guarantees the best HA characteristics.
```
l-mb: But it doesn't :-) The need to synchronize and coordinate all operations across all nodes (to guarantee Atomicity/Independence/Durability, which even S3's eventual consistency model needs in some cases) is not free - even then there's a need to ensure the data has been replicated in reality, and since a fault can only be detected via a timeout of some form, there's still a blip in availability (just, hopefully, not a returned error message).

And since the synchronization mechanism is often complex, there's an on-going price to be paid for achieving this, plus it is harder to get right (lowering availability through lowered quality).

I'll argue that *for our use case*, the active/standby model built on top of Longhorn actually makes the most sense and brings the "best" HA characterics relative to implementing a more fully active/distributed solution. (And if someone wanted it, it already exists, it's called Ceph/RADOS.)

docs/research/ha/RATIONALE.md
```diff
@@ -0,0 +1,446 @@
...
+where all the pieces are replicated.
+
+The immediate consequence of this statement is that every pipeline being part of
+the same logical s3gw service must bind to a different `PV`; one per `radosgw` process.
+All the `PV`s for the same logical s3gw, must therefore, sync their data.
+
+This implies that with this model, the s3gw must implement an efficient replication
+model across all the pipelines being part of the same logical service.
+
+Under these conditions, a client accessing the S3 service would experience a failure
```
l-mb: No, it might not mean that; the ingress might still mask it and automatically retry. (That's another thing to consider: most S3 protocol libraries are built on the assumption of an unreliable network, and so a single non-repeatable failure might not show to the user, but just be (silently or not) retried and on it goes without anyone knowing beyond a blip in high latency, unless it happens too often. This is a property of the S3 protocol that makes it a bit easier for us to achieve, we hope.)

docs/research/ha/RATIONALE.md
```diff
@@ -0,0 +1,446 @@
...
+
+With pipeline we mean all the chain from the ingress to the `PV`:
+
+```mermaid
+flowchart LR
+    ingress(ingress) ==> radosgw(radosgw) ==> PV((PV))
+```
+
+The difficulty with a project like the s3gw is due to the fact that basically
+we have one process: the `radosgw`, that has the *exclusive* access to one resource:
```
l-mb: When we look at the Single Points of Failure in our stack, another one is the ingress, likely. We'll have to consider that in addition to the s3gw pod. Or at least explain why we consider it out of scope.

docs/research/ha/RATIONALE.md
```diff
@@ -0,0 +1,446 @@
...
+
+## Active-Passive pipelines
+
+This model assumes N pipelines to be "allocated" on the cluster at the same time,
+but only one, the *active* pipeline, owns the *exclusive ownership*
+over the shared `PV`.
+
+Should the active pipeline suffer a failure, the next elected active pipeline,
+chosen to replace the old one, should pay a *time window* necessary to "transfer"
+the ownership over the shared `PV`.
```
l-mb, line 138: As explained above, there's also a latency to active/active setups to detect and recover from faults.

The active plus warm standby operation here has no meaningful advantages over the passive/fail-over option; loading the s3gw pod is the cheapest part of the whole process, compared to fault detection (usually a timeout), mounting (journal recovery) of the file system, the process running the SQLite etc recovery on start, etc.

We'd pay for this with complexity (and resource consumption while in standby) that likely would only give us very marginal benefits at best. I'd argue that beyond explaining why, we do not need to further investigate this option.

docs/research/ha/RATIONALE.md
```diff
@@ -0,0 +1,446 @@
...
+    linkStyle 2 stroke: green
+    linkStyle 3 stroke: green
+```
+
+## One Active pipeline
+
+This model is logically equivalent to the *Active-Passive* model with the difference
+that only one pipeline is allocated on the cluster at a time.
+
+In the *One pipeline* model, the only notable perceived difference between
```
l-mb: Again, I argue this isn't true. The fail-over latency is very unlikely to be measurably worse (there's nothing meaningful preloaded in the second model), while at the same time, the fault-free operation performance is going to be better, and the complexity is much lower.

docs/research/ha/RATIONALE.md
```diff
@@ -0,0 +1,446 @@
...
+the *Active-Passive* model are the worst performances obtained in the event of failure.
+
+In this scenario, the system should pay also the time needed to fully load a new
+pipeline, that unlike that *Active-Passive* model, is not preloaded.
+
+So, supposing that the new pipeline has scheduled to load on a node where the
```
l-mb: Now this shouldn't happen, good point! (I keep forgetting about it because I still grew up in a world where all software was installed locally :-) ) We need to make sure all our images are preloaded on the nodes.
Downloading the image from the registry at this time wouldn't be acceptable.

docs/research/ha/RATIONALE.md
```diff
@@ -0,0 +1,446 @@
...
+
+The 3 models described above have different performances and different implementation
+efforts. While the *Active-Active* model is expected to require a significant
+development effort due to its inherent complex nature, the *Passive-Active* model
+seems to offer a lower degree of implementation difficulty.
+
+In particular, the *One Active pipeline* model, is expected to work with nothing
+but the primitives normally available on any Kubernetes cluster.
+
+Given that the backend s3gw architecture is composed by:
```
l-mb, line 213: Should we add the ingress here? I am rather sure that we need to consider it in our model.

(Achieving an HA ingress may be out of scope, since that just kicks up the problem up one layer, but I suspect that while the ingress is functional, it can help us mask faults.)

docs/research/ha/RATIONALE.md
```diff
@@ -0,0 +1,446 @@
...
+
+- one *stateless* `radosgw` process associated with a *stateful* `PV`
+
+it is supposed that, on `radosgw` process failure, a simple reschedule
```
l-mb: Yes and no. k8s has policies that trigger a local restart (via monitoring/probes), but can also trigger a pod to go to another node. All of this has timeouts and delays, and I suspect we'll have to adjust them for our requirements.

docs/research/ha/RATIONALE.md
```diff
@@ -0,0 +1,446 @@
...
+of its POD is enough to fulfill this simple HA model.
+A clarification is needed here: the `PV` *state* is guaranteed by a third party
+service, that is: **Longhorn**.
+
+Longhorn ensures that a `PV` (along with its content) is always available on
```
l-mb, line 225: While we consider failure modes, this is only true if the cluster still has quorum and there may be LH options that affect availability and recovery. It's also important to call out that obviously we can't achieve higher availability or reliability than the underlying LH volumes.

docs/research/ha/RATIONALE.md
```diff
@@ -0,0 +1,446 @@
...
+any cluster's node, so that, a POD can mount it regardless of its allocation on the
+cluster.
+
+Relying on this assumption, while we are still in the process to decide what model
```
l-mb: Strongly suggest we ignore everything except the active/standby "one pipeline" model. Everything else is too complicated for, at the _very_ least, the first iteration. We'd have to significantly rearchitect too much of the stack, it's incredibly difficult to explain to users, and one of the key goals of LH/s3gw together is also simplicity.

Let's cut to the chase and pick this model for investigation, design, and implementation ASAP.

docs/research/ha/RATIONALE.md
```diff
@@ -0,0 +1,446 @@
...
+we want to actually engage, the *One Active pipeline* model seems for sure
+the easiest thing to explore and to investigate.
+
+## Failure cases
+
+The `PV` state is kept coherent by Longhorn, so errors at this level are assumed
+NOT possible (application level corruptions to the `PV`'s state ARE possible).
```
l-mb: Not quite; s3gw won't corrupt the PV's state or the file system on it, but it _might_ corrupt its own application data. (Sorry for splitting that hair, but it's an important distinction: any corruption in the file system for example is outside what s3gw can reasonably protect against; at best, we can crash in a way that doesn't corrupt our data further, but that's all undefined behaviour and "restore from backup" time.)

docs/research/ha/RATIONALE.md
```diff
@@ -0,0 +1,446 @@
...
+so that they trigger an anomalous exit.
+
+### radosgw's POD failure
+
+This case is when the `radosgw` process stops due to a failure of its POD.
+This case applies also when Kubernetes decides to reschedule the POD
+to another node, eg: when the node goes low on resources.
+
+This can be thought as an infrastructure issue independent to the s3gw.
+In this case, the *One Active pipeline* model fully restores the service by
```
l-mb: This would often be called a "switch-over" in an HA cluster - e.g., an administratively orchestrated transition of the service to a new node (say, for maintenance/upgrade/etc reasons).
This has the advantage of being schedulable, so it can happen at times of low load if these exist.

But in combination with proper interaction with the ingress - pausing requests there instead of failing them - we should be able to mask these cleanly.

Bonus: this is also what we need to seamlessly restart on upgrade/update of the pod itself transparently.

This is 100% within our scope, I'd say.

docs/research/ha/RATIONALE.md
```diff
@@ -0,0 +1,446 @@
...
+rescheduling a new POD somewhere in the cluster.
+
+### Cluster's node failure
+
+This case is when the `radosgw` process stops due to a cluster's node failure.
+The *One Active pipeline* model fully restores the service by
```
l-mb: Right, but this also means we weren't shut down cleanly. So the on start recovery needs to be optimized for the stack, and as soon as we can, we need to hook into the ingress and tell it to pause until we're done, and then resume.

docs/research/ha/RATIONALE.md
```diff
@@ -0,0 +1,446 @@
...
+rescheduling a new POD somewhere in the cluster.
+
+### radosgw's failure due to a bug not related to a certain input pattern
+
+This case is when the `radosgw` process crashes due to a bug not directly
```
l-mb: Does it actually matter *why* the process crashed? I think this will be very difficult to differentiate in practice - a query can trigger a memory leak, it can just accumulate and get worse over time, etc.
I think this and the next case can just be combined together as abnormal process termination.

docs/research/ha/RATIONALE.md
```diff
@@ -0,0 +1,446 @@
...
+the user's operations under the condition that the crash-triggering
+input is recognized by the user and thus its submission blocked.
+
+### PV Data corruption at application level due to radosgw's bug
+
+This case is when the state on the `PV` corrupts due to a `radosgw`
+bug.
+
+In this unfortunate scenario, the *One Active pipeline* can hardly help.
+A restart could evenly fix the problem or trigger an endless restarting loop.
```
l-mb, line 342: Right, none of the models can help here. Logical data corruption is a Robustness/Reliability problem; the best we can aim for is to detect it (and abort with prejudice and finally so, so as to not make the corruption worse).
We can't even tell it apart from data corruption at the actual PV level; all we know the data isn't consistent/coherent.

I think beyond that ("don't make it worse"), this case is outside the scope of an HA scenario. There's no way to recover here. This is more related to Tim's work on the s3gw.fsck

docs/research/ha/RATIONALE.md
```diff
@@ -0,0 +1,446 @@
...
+
+The fix for this could contemplate an human intervention.
+A definitive solution would be available only when a patch for the issue
+is available.
+
+### PV Data corruption at application level due to radosgw's anomalous exit
+
+This case is when the state on the PV corrupts due to a `radosgw`
```
l-mb: This case is functionally identical to the previous one, we have no way of knowing where or how the corruption was introduced

docs/research/ha/RATIONALE.md
```diff
@@ -0,0 +1,446 @@
...
+
+After reviewing the cases, we can say that what can be actually solved with
+an HA model with s3gw is when the failure is not dependent to applicative bugs.
+We can handle temporary issues involving the infrastructure that is hosting
+the `radosgw` process.
+
+So, we are mostly interested in observing and measuring the timing of
+Kubernetes when a `radosgw`'s POD is being rescheduled.
+
+Hence, The idea to collect measures regarding a series of restarts
```
l-mb, line 378: I think we need separate things.

The first is the (kill - re)start loop timing outside of k8s/LH (so we have a baseline and can measure of what k8s adds, and how slow we are when exiting cleanly, crashing with no ops in flight, vs a bazillion point three five parallel ops on-going).

Then, fault detection times for k8s - how long until it notices that the process has crashed (that should be quick due to the generated signal), but what about the process *hanging*? (e.g., crashes are separate from timeouts)

For node failures, again, we need to understand which factors affect k8s detecting that and reacting to them, and what latencies are introduced by k8s/LH.

Actively asking k8s to restart is different (see case one, switch- vs fail-over). That should be smooth, but is not actually a failure scenario. Probably worth handling in a separate section.

docs/research/ha/RATIONALE.md
```diff
@@ -0,0 +1,446 @@
...
+artificially triggered on the `radosgw`'s POD.
+Obtaining such measures would allow to compute some arbitrary statistics
+for the time required by Kubernetes to restart the `radosgw`'s POD.
+
+## The s3gw Probe
+
+The `s3gw Probe` is a program developed with the purpose of collecting restart
```
l-mb: This strikes me as debug-/test-level operations that might fit in well with the status/admin extensions that Marcel is adding to the s3gw process for the metrics/status/monitoring page?

And the logs should be available from k8s.

Do we need a separate pod for this or can we fold it into our existing processes?

(Testing this ultimately end-to-end will require integration with the LH test suite, if they have automated fault injection testing and recovery timing; that may be far-fetched for the evaluation of this eventual ADR, but the design the ADR should propose to continuously test/validate our HA concept should tie into that framework.)

l-mb:
A high-level comment would be to touch base with the LH team on the work on the NFS share manager and their HA plans/ideas. While they can't take advantage of an ingress, they share a few of the similar concerns - node/pod failure detection, recovery, etc.

Perhaps there's overlap and tech we can leverage jointly.

docs/research/ha/RATIONALE.md
giubacc: > I'd argue our key goal is to maximize user-visible, reliable service availability, and to be tolerant with regard to certain faults

I Agree on the goal's high level idea; we have to translate this statement in a precise requisite that can be then implemented.
I'd say also that the HA model could be very much dependent to the specific need of the user; for a certain workload could be more important:

- After have experienced a failure, have the guarantee that the S3 service will be available after X seconds. 
- After have experienced a failure, have the guarantee that the S3 service will be available on the NEXT (+X) access. (where X is something that goes from 0..N)

docs/research/ha/RATIONALE.md
giubacc: Good arguments :) . I agree.

docs/research/ha/RATIONALE.md
giubacc: Right, I will add this consideration.

docs/research/ha/RATIONALE.md
giubacc: roger :+1:

docs/research/ha/RATIONALE.md
giubacc: yes, it seems absolutely reasonable to add it to the model.

docs/research/ha/RATIONALE.md
giubacc: What do you mean exactly with  _if the cluster still has quorum_ ?

> It's also important to call out that obviously we can't achieve higher availability or reliability than the underlying LH volumes.

I'd say that the s3gw has more to do with the concept of S3 _availability_ while LH has the responsibility to maintain the state of the underlying data in case of severe failure.
We could even say that our model  prefers data preservation over the (S3) data _availability_ and thus that in certain cases it is preferable to completely stop the  _availability_ to  maximize the preservation of the current  data state.
That is, we could prefer the lesser bad (stop the access to S3 clients)  to safeguard the data state.
(I'm just thinking aloud here)

docs/research/ha/RATIONALE.md
giubacc: It seems reasonable. :+1:

docs/research/ha/RATIONALE.md
giubacc: I think we are saying the same thing :)
All that is _ABOVE_ the filesystem can be corrupted by the s3gw application level.

docs/research/ha/RATIONALE.md
giubacc: About this, should we track somewhere in the sqlite database that the `radosgw` process has cleanly exited?
When the `radosgw` starts, it could perform some sanity check over the data before (eventually) accepting new client's requests. In certain cases we could decide that the state is not recoverable and a manual intervention is needed.

docs/research/ha/RATIONALE.md
giubacc: I think that make these cases explicit is useful for a theoretical reasoning on what scenarios we can actually think to solve with an HA model.
This case is clearly outside what an HA model can handle; we must expect for this case that the `back off` mechanism is the only mitigation when a restart loop should occur.

docs/research/ha/RATIONALE.md
giubacc: This so called s3gw-probe, has been developed to ease the collecting of stop/start events for this specific research (and potentially to compute some statistics when and if needed) ; it was not meant (at least it wasn't my initial though) to make an official tool with a more authoritative role outside the scope of this research. 
If we want to make all the tests and measures we will do for this research officially replicable, I agree that we must devote a coordination effort to understand where to put all the probes and what kind of sources do we want to use to collect the events (e.g. the logs).

docs/research/ha/RATIONALE.md
m-ildefons: In general on K8s, [it can not be assumed that an image is pre-pulled](https://kubernetes.io/docs/concepts/containers/images/#pre-pulled-images), since nodes may come and go dynamically.

james-munson:
One item we (Longhorn) would like to know about the S3GW HA is whether it will assume the volume beneath the object store must be RWX, or whether RWO would suffice.  If the gateway is active/active enough that both sides need simultaneous write access in order to transfer the work fast enough, that will make a difference.  

From what I gather in the discussion here, it is unacceptable to have to start a pod on the new owner as part of the failover, but should be acceptable to defer attaching the backing volume until then.  If so, then a simple RWO volume would suffice.

If not, the RWX volume would itself be layered on NFS, and any HA transfer would be gated by the NFS HA transfer, which currently requires significant time to clear locks, wait for grace periods, and all that.  (I have the ticket to try to improve its performance, if possible.)

giubacc:
> One item we (Longhorn) would like to know about the S3GW HA is whether it will assume the volume beneath the object store must be RWX, or whether RWO would suffice. If the gateway is active/active enough that both sides need simultaneous write access in order to transfer the work fast enough, that will make a difference.
> 
> From what I gather in the discussion here, it is unacceptable to have to start a pod on the new owner as part of the failover, but should be acceptable to defer attaching the backing volume until then. If so, then a simple RWO volume would suffice.
> 
> If not, the RWX volume would itself be layered on NFS, and any HA transfer would be gated by the NFS HA transfer, which currently requires significant time to clear locks, wait for grace periods, and all that. (I have the ticket to try to improve its performance, if possible.)

Our current idea is to propose the HA model: "active/standby".
This can be translated into Kubernetes concepts as: a deployment with an "immutable" `replicas: 1` spec.
So in case of failure, Kubernetes would restart a new s3gw's pod that will gain the "ownership" over the LH volume without too much complications; we suppose it would be a _not-failable_ operation for the new s3gw's pod being able to attach to the existing LH PVC (previously mounted on the failed pod).  
In this way ("active/standby"), we would avoid to have others "cold" instances of the s3gw; an "active/passive" approach would require a potentially not trivial complexity to be added over the Kubernetes primitives we should implement.

docs/research/ha/RATIONALE.md
l-mb: Right. Since we don't know how many ops/s the user pushes, and our recovery time is (mostly) constant (possibly related to the number of ops we need to replay from the journal etc?), I think we'd go with the "Mean Time to Repair" or more likely "Recovery Time Objective" in response to the failure.

That also allows us to specify the "Recovery *Point* Objective" - which operations are possibly lost? e.g., in-flight operations at the time of most faults will have to be retried by the client side (we can't resume the TCP/IP connection).

docs/research/ha/RATIONALE.md
l-mb: This is a great point even then, because this is something we *do* want to ensure for eligible nodes. Otherwise, our restart is unpredictably long.

(It's always possible that a fault occurs exactly at the time where we are pre-loading the image, but that's just bad luck.)

This also matters for updates, I assume.

docs/research/ha/RATIONALE.md
l-mb: I think you could configure LH in a way that would allow both sides to proceed in case of a network partition (which would lead to irreconcilable divergence on the PV). We should call out that these "split brain scenarios" must not occur.

And indeed, `s3gw` should not spin up if it detects a logically corrupt store. I'm guessing that's kind of implied since I assume it'd crash in that case :-D And this will also never be perfect, but it might be worth considering if we can.

docs/research/ha/RATIONALE.md
l-mb: Here's a fun idea to ponder based on previous experience with some databases - because they tried to exit "cleanly", the shutdown took so long that they were faster to just crash and force into the recovery/journal replay on start. :-D

I think that a clean shutdown will surely be beneficial, since it would flush the WAL etc, but I don't think we should record it. If we have start-up checks, we should always execute them (since, see further above, what we're trying to detect might be a logical error that slipped past). For everything else, we rely on the consistency guarantees of sqlite and the file system.

docs/research/ha/RATIONALE.md
l-mb: Fair, we can rely on the back-off loop in some cases (though _maybe_ the admin wants to disable that so that, once the possibly faulty-itself client stops, s3gw restarts as quickly as possible).

My initial suggestion would be to focus on node crashes (or their equivalent, such as OOM etc), and restarts due to that, since that's also what clients are most likely to test.

docs/research/ha/RATIONALE.md
l-mb: Right, ultimately I think we want to somehow integrate this with the longhorn-tests, I suspect, if that's feasible at all, so it can truly be tested end-to-end at the system level.

l-mb:
> One item we (Longhorn) would like to know about the S3GW HA is whether it will assume the volume beneath the object store must be RWX, or whether RWO would suffice. If the gateway is active/active enough that both sides need simultaneous write access in order to transfer the work fast enough, that will make a difference.

`RWO` is, in fact, the only supported mode.

We need features from XFS that NFS would no longer expose; and we will not support multiple `s3gw` instances on the same store (which conceptually also doesn't really make sense from a performance PoV, and not really from an availability point of view either, since it still all depends on a single node).

We've got no plans to support `RWX`.

(At that point, s3gw would be slowly implementing a distributed K/V object store as a backend, and ... that'd be called RADOS/Ceph :-D )

docs/research/ha/RATIONALE.md
m-ildefons: > though maybe the admin wants to disable that so that, once the possibly faulty-itself client stops, s3gw restarts as quickly as possible

afaik, the backoff timer can not be disabled (or tuned) without modifying K8s code.

docs/research/ha/RATIONALE.md
l-mb: With a little more time, it seems that "RTO" is more common in the storage and DR/BC worlds. (MTTR is more often used to repairing the underlying cause.)

It also fits better with our "RPO" concept (which should be zero for completed and acknowledged operations, in-flight and not acknowledged ones might need to be restarted by the client).

jecluis:
@l-mb are we good to merge this?

jecluis:
@giubacc there  are conflicts with this PR, mind addressing them?

giubacc:
rebased on latest main

giubacc:
@l-mb @jecluis @vmoutoussamy 
Can we merge this first HA research?
I'd rather handle the current activity over medik8s with its [dedicated LH issue.](https://github.com/longhorn/longhorn/issues/6803)

jecluis (approved):

PR #697, opened 2023-09-01
Add decision about versioning of the UI REST API

votdev:
# Describe your changes

### DIFF
--- a/docs/decisions/0014-ui-backend-api.md
+++ b/docs/decisions/0014-ui-backend-api.md
@@ -49,6 +49,20 @@ Therefore, we have chosen to use POST operations for selected operations that
 are semantically the equivalent to a GET operation. This breaks the RESTfulness
 of our API for some operations, but we believe this to be a reasonable tradeoff.
 
+## Versioning
+
+Versioning of an API makes sense if it is public, but the UI backend API is
+not meant for public use. Since the UI and the associated backend REST API
+are always delivered together and are therefore tightly coupled, there is no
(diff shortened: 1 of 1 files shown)


jecluis (dismissed):
sounds good to me.

irq0:
What if a browser has the UI open with all the javascript loaded and the backend restarts to a new version. Will the UI app be able to detect that without any versioning information?

jecluis:
> What if a browser has the UI open with all the javascript loaded and the backend restarts to a new version. Will the UI app be able to detect that without any versioning information?

That can be prevented with a version check, but doesn't necessarily mean maintaining API versions. One can increase the API version if things change or are incompatible, and share it with the UI as an HTTP header or something, and then have the UI check whether that's the version it knows -- if not, ask the user to refresh.

I think the change to the ADR still makes sense though, it makes little sense to introduce proper API versioning IMO.

irq0 (dismissed):
sgtm

Unimportant nit: Haven't seen 'S3gw' before, only 'S3GW' or 's3gw'

votdev:
> Unimportant nit: Haven't seen 'S3gw' before, only 'S3GW' or 's3gw'

That's the result of using camel-case. I've changed it to `S3GW-Foo-Bar`.

BTW, using the `X-` prefix in custom headers is [deprecated](https://www.rfc-editor.org/rfc/rfc6648).

jecluis:
lgtm

jecluis (approved):

PR #707, opened 2023-09-12
Add Metrics Documentation

irq0:


jecluis (approved):
commit needs to be signed, but otherwise lgtm

PR #729, opened 2023-09-28
Dockerfile: fix sqlite version to 3.43.1

### LINKED ISSUE #725: s3gw: fix sqlite version (maybe others?) on Dockerfile
Fixing versions in the Dockerfile will allow us to upgrade packages without them affecting already released versions, or on-going releases.

This comes about via #723 where we kind of want to upgrade sqlite, but because we're not fixing the version in the Dockerfile, upgrading it right now in https://build.opensuse.org/project/show/filesystems:ceph:s3gw would affect the on-going release effort, upgrading a major library right from under us while the release process is on-going.

Additionally, by not fixing releases in the Dockerfile, we fall into the trap of having, e.g., v0.21.0 on sqlite X.Y.A, and then all versions post v0.21.0, including v0.21.1, v0.21.2, etc (should they exist), on a future version of sqlite X.Y.B (or something similar).

jecluis:
To be merged after @tserong updates the  packages in https://build.opensuse.org/project/show/filesystems:ceph:s3gw and adds the newest sqlite3 3.43.1 version.

Fixes: aquarist-labs/s3gw#725

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,6 +6,7 @@ RUN zypper ar \
  && zypper --gpg-auto-import-keys ref
 
 RUN zypper -n install \
+  libsqlite3-0=3.43.1 \
   libblkid1 \
   libexpat1 \
   libtcmalloc4 \


tserong (approved):
LGTM, but probably also wants to be done in tools/build/Dockerfile.build-radosgw and tools/build/Dockerfile.s3gw (although the latter should arguably be ultimately removed in favour of the top-level Dockerfile, but that can be done separately - see https://github.com/aquarist-labs/s3gw/issues/731)

jecluis:
Yeah, I think doing it as part of #731 makes sense :)

tserong (changes requested):
I've got 3.43.1 into filesystems:ceph:s3gw now, tried a build, and we need one more little tweak here:

```diff
diff --git a/Dockerfile b/Dockerfile
index 5ef048b..1fadc39 100644
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,7 @@ RUN zypper ar \
   s3gw-deps \
  && zypper --gpg-auto-import-keys ref
 
-RUN zypper -n install \
+RUN zypper -n install --allow-vendor-change \
   libblkid1 \
   libexpat1 \
   libtcmalloc4 \
```

The problem is that the base image (opensuse/leap:15.4) already has libsqlite3-0-3.39.3-150000.3.20.1.x86_64 installed, so trying to install sqlite-devel=3.43.1 triggers this rotten thing:

```
156.6 Resolving package dependencies...
156.7 
156.7 Problem: the to be installed sqlite3-devel-3.43.1-lp154.322.1.x86_64 requires 'libsqlite3-0 = 3.43.1', but this requirement cannot be provided
156.7   not installable providers: libsqlite3-0-3.43.1-lp154.322.1.x86_64[s3gw-deps]
156.7  Solution 1: install libsqlite3-0-3.43.1-lp154.322.1.x86_64 from vendor obs://build.opensuse.org/filesystems
156.7   replacing libsqlite3-0-3.39.3-150000.3.20.1.x86_64 from vendor SUSE LLC <https://www.suse.com/>
156.7  Solution 2: do not ask to install a solvable providing sqlite-devel = 3.43.1
156.7  Solution 3: break sqlite3-devel-3.43.1-lp154.322.1.x86_64 by ignoring some of its dependencies
156.7 
156.7 Choose from above solutions by number or cancel [1/2/3/c/d/?] (c): c
```

tserong:
Wait a minute, it doesn't quite work even with that change. Stay tuned...

tserong:
...uh, the previous suggested tweak was in the wrong place >facepalm<. You actually want this one:

```diff
diff --git a/Dockerfile b/Dockerfile
index 5ef048b..491cc0b 100644
--- a/Dockerfile
+++ b/Dockerfile
@@ -43,7 +43,7 @@ ENV SRC_CEPH_DIR="${SRC_CEPH_DIR:-"./ceph"}"
 ENV ENABLE_GIT_VERSION=OFF
 
 # Add OBS repository for additional dependencies necessary on Leap 15.4
-RUN zypper -n install --no-recommends \
+RUN zypper -n install --no-recommends --allow-vendor-change \
       'cmake>3.5' \
       'fmt-devel>=6.2.1' \
       'gperftools-devel>=2.4' \
```

(Just waiting for a build to complete now, but I got past the install step with the above patch).

tserong:
Nope, that's not it either. The above works fine to get the right version of sqlite-devel installed when _building_ s3gw, but the container that s3gw _runs_ in still has version 3.39.3.

This is rapidly turning into a textbook example of why:
a) reviewers should always test dependency changes completely before giving them the OK
-- and --
b) one should never do reviews on a Friday night.

tserong:
Alright, _this_ does it:

```diff
diff --git a/Dockerfile b/Dockerfile
index 5ef048b..c1b9c19 100644
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,6 +6,7 @@ RUN zypper ar \
  && zypper --gpg-auto-import-keys ref
 
 RUN zypper -n install \
+  libsqlite3-0=3.43.1 \
   libblkid1 \
   libexpat1 \
   libtcmalloc4 \
```

That's installing the desired version of libsqlite3-0  in s3gw-base, which is used by all the other containers. When buildenv is created, it pulls in the sqlite-devel version which matches the already installed libsqlite3-0 (so we don't need to say sqlite-devel=3.43.1, it just works without the version there) . Apparently there's no need for `--allow-vendor-change` when doing it this way.

tserong (approved):

PR #733, opened 2023-10-02
Add decision about the HA model to be used with the s3gw with Longhorn

giubacc:
# Describe your changes

Add decision about the HA model to be used with the s3gw with Longhorn

Signed-off-by: Giuseppe Baccini <giuseppe.baccini@suse.com>

## Issue ticket number and link

Related to: https://github.com/aquarist-labs/s3gw/issues/361

### DIFF
--- a/docs/decisions/0018-s3gw-ha-model.md
+++ b/docs/decisions/0018-s3gw-ha-model.md
@@ -1,0 +1,41 @@
+# s3gw High Availability model
+
+## Context and Problem Statement
+
+We analyzed some High Availability - HA - concepts applied to the s3gw when used with Longhorn.
+The final aim of the research is to identify an HA model we can reasonably rely on.
+
+The full and initial HA research document can be found here:
+[High Availability research](../research/ha/RATIONALE.md).
+You can find there all the rationales, motivations and the details about the tests performed.
+
+## Considered Options
+
+We identified 3 HA models:
+
+- **Active/Active** (multiple s3gw instances concurrently serving the same data)
+- **Active/Warm Standby** (multiple s3gw instances, one serving data, others able to take over if active instance fails)
+- **Active/Standby** (single s3gw instance, with Kubernetes restarting/redeploying as necessary on failure)
+
+## Decision Outcome
+
+The 3 aforementioned models have different performances and different implementation efforts.
+For our use case, the *Active/Standby* model built on top of Longhorn actually makes
+the most sense and brings the "best" HA characteristics relative to implementing a
+more fully active/distributed solution.
+
+List of *desirable* characteristics owned by the *Active/Standby* model
+
+- Simplicity
+- Low implementation effort in respect to the other models
+- Expected to work mainly with Kubernetes primitives
+- Compatible with RWO persistent volume semantics
+- Acceptable restart timings on switch-overs and fail-overs (excluding the non-graceful node failure)
+
+Be aware that the [non-graceful node failure](../research/ha/RATIONALE.md#non-graceful-node-failure)
+problem cannot be entirely solved with the *Active/Standby* model alone.
+Regarding this, we have opened a [dedicated issue](https://github.com/longhorn/longhorn/issues/6803)
+within the Longhorn project.
+
+For a more comprehensive explanation about this choice, please refer to the original
+[High Availability research](https://github.com/aquarist-labs/s3gw/pull/685) pull request.


tserong (commented):
I know the full details are in the research docs, but I wonder if we should add a very brief sentence to provide a little more detail about each mode? Maybe something like this (assuming I've managed to capture the essence):

- Active/Active (multiple s3gw instances concurrently serving the same data)
- Active/Warm Standby (multiple s3gw instances, one serving data, others able to take over if active instance fails)
- Active/Standby (single s3gw instance, with Kubernetes restarting/redeploying as necessary on failure)

Otherwise, LGTM. I assume we'll update the PR links later to point to the research docs once that PR is merged.

giubacc:
> I know the full details are in the research docs, but I wonder if we should add a very brief sentence to provide a little more detail about each mode? Maybe something like this (assuming I've managed to capture the essence):
> 
> * Active/Active (multiple s3gw instances concurrently serving the same data)
> * Active/Warm Standby (multiple s3gw instances, one serving data, others able to take over if active instance fails)
> * Active/Standby (single s3gw instance, with Kubernetes restarting/redeploying as necessary on failure)
> 
> Otherwise, LGTM. I assume we'll update the PR links later to point to the research docs once that PR is merged.

that makes perfectly sense, will change; moreover, pointing the PR is something not ideal probably, it will make more sense point the HA document directly once the corresponding PR will be merged.

docs/decisions/0015-s3gw-ha-model.md
```diff
@@ -0,0 +1,31 @@
...
+The 3 HA models have different performances and different implementation efforts.
+For our use case, the *Active/Standby* model built on top of Longhorn actually makes
+the most sense and brings the "best" HA characteristics relative to implementing a
+more fully active/distributed solution.
+
+The *Active/Standby* model is expected to work with nothing but the primitives normally
+available on any Kubernetes cluster.
+
+for a more comprehensive explanation about this choice, please refer to the
+[High Availability research](https://github.com/aquarist-labs/s3gw/pull/685) Pull request.
```
jecluis: for consistency, I'd either capitalize "Request", or write "pull request" instead.

docs/decisions/0015-s3gw-ha-model.md
```diff
@@ -0,0 +1,31 @@
...
+
+The 3 HA models have different performances and different implementation efforts.
+For our use case, the *Active/Standby* model built on top of Longhorn actually makes
+the most sense and brings the "best" HA characteristics relative to implementing a
+more fully active/distributed solution.
+
+The *Active/Standby* model is expected to work with nothing but the primitives normally
+available on any Kubernetes cluster.
+
+for a more comprehensive explanation about this choice, please refer to the
```
jecluis: capitalize "For"

docs/decisions/0015-s3gw-ha-model.md
```diff
@@ -0,0 +1,31 @@
...
+
+- *Active/Active*
+- *Active/Warm Standby*
+- *Active/Standby*
+
+## Decision Outcome
+
+The 3 HA models have different performances and different implementation efforts.
+For our use case, the *Active/Standby* model built on top of Longhorn actually makes
+the most sense and brings the "best" HA characteristics relative to implementing a
```
jecluis: would be nice to have a small bullet point list of what are the perceived "best" HA characteristics.

docs/decisions/0015-s3gw-ha-model.md
```diff
@@ -0,0 +1,31 @@
+# s3gw ha model
+
+## Context and Problem Statement
+
+We analyzed some High Availability - HA - concepts applied to s3gw when used with Longhorn.
+The final aim of the research is to identify an HA model we can use with the s3gw.
+
+The full HA research work conducted until now has been proposed with the
+[High Availability research](https://github.com/aquarist-labs/s3gw/pull/685) Pull request.
```
jecluis: either "Pull Request" or "pull request". I'm partial towards the latter.

jecluis (changes requested):

docs/decisions/0015-s3gw-ha-model.md
```diff
@@ -0,0 +1,31 @@
+# s3gw ha model
```
jecluis: and please capitalize "HA" here as well :)

jecluis: I'd even go further and actually write "High Availability" instead.

docs/decisions/0015-s3gw-ha-model.md
giubacc: For the time being I will go with all downcase; anyway, I'm not sure we should point the PR; perhaps better link the HA document once merged?

docs/decisions/0015-s3gw-ha-model.md
giubacc: For the time being I will go with all downcase; anyway, I'm not sure we should point the PR; perhaps better link the HA document once merged?

l-mb:
A key part that we also need to address in our stack is the ingress layer, since we can hide many problems behind a proxy that stalls the clients, and the time of making it connect to a new backend affects our timings as well.

giubacc:
> A key part that we also need to address in our stack is the ingress layer, since we can hide many problems behind a proxy that stalls the clients, and the time of making it connect to a new backend affects our timings as well.

Currently, with Traefik, we are getting a "NotFound: Not Found\n\tstatus code: 404, request id: , host id: " when the s3gw backend service is down.
Could be better to address Ingress specific investigations in a separate thread?
Maybe we can write in this ADR that the Ingress part needs a specific investigation effort and we reserve to do this with a separate dedicated issue/epic?

irq0:
> Currently, with Traefik, we are getting a "NotFound: Not Found\n\tstatus code: 404, request id: , host id: " when the s3gw backend service is down.

That is not useful - clients won't retry a 404. 503 Service Unavailable would be a good option as clients like boto3 retry them https://github.com/boto/boto3/blob/develop/docs/source/guide/retries.rst

l-mb:
Basically, if we push the ingress out - just like node failure recovery, or pre-loading the images - the actual ADR boils down to "s3gw needs to be able to be crash-consistent and start up fast after such an outage, because our active/cold standby model relies on the framework around us to achieve HA."

Then, this ADR would set a worst-case recovery goal (crashed under heavy load, say) of <1-5s for the s3gw restart part of the full stack.

That's fair (and mostly true), but then we need to immediately start on the next LEP to discuss how to achieve HA goals (RTO < 30s or some such goal) end-to-end. (And how to measure/validate those numbers as part of our E2E system testing, perhaps using k6?)

giubacc:
> Basically, if we push the ingress out - just like node failure recovery, or pre-loading the images - the actual ADR boils down to "s3gw needs to be able to be crash-consistent and start up fast after such an outage, because our active/cold standby model relies on the framework around us to achieve HA."
> 
> Then, this ADR would set a worst-case recovery goal (crashed under heavy load, say) of <1-5s for the s3gw restart part of the full stack.
> 
> That's fair (and mostly true), but then we need to immediately start on the next LEP to discuss how to achieve HA goals (RTO < 30s or some such goal) end-to-end. (And how to measure/validate those numbers as part of our E2E system testing, perhaps using k6?)

I have the impression/feeling that the activity on the ingress thing (testing, researching modes that fit us, etc) is something that needs a proper allocation in terms of time at management level. That is something that must be agreed, especially now we are more committed in integrating with LH.

jecluis:
@l-mb is there anything else you'd like to see done here?

docs/decisions/0015-s3gw-ha-model.md
jecluis, line 1: file name needs to be changed. Decision 0015 has been used.

jecluis (changes requested):

docs/decisions/0015-s3gw-ha-model.md
jecluis: I suggest using 0018 to avoid conflicting with other in-flight pull requests

giubacc:
@jecluis @l-mb ping

docs/decisions/0018-s3gw-ha-model.md
```diff
@@ -0,0 +1,39 @@
...
+
+List of *desirable* characteristics owned by the *Active/Standby* model
+
+- Simplicity
+- Low implementation effort in respect to the other models
+- Expected to work mainly with Kubernetes primitives
+- Compatible with RWO persistent volume semantics
+- Acceptable restart timings on switch-overs and fail-overs (excluding the non-graceful node failure)
+
+Be aware that the [non-graceful node failure](https://github.com/aquarist-labs/s3gw/blob/4af657c573ce634cd16c53c20986e54817077b44/docs/research/ha/RATIONALE.md#non-graceful-node-failure) problem cannot be entirely solved with the *Active/Standby* model alone.
```
jecluis: no idea how this is passing linting. Regardless, mind fixing this line to a proper column size? Maybe move the link to a ref at the end of the document to mitigate the noise?

jecluis (changes requested):

docs/decisions/0018-s3gw-ha-model.md
giubacc: I've instead linked the now available HA document directly with its natural project path.
That is much more short and it looks definitely better.

jecluis (approved):

PR #735, opened 2023-10-03
docs/decisions: add s3gw development mode ADR

giubacc:
# Describe your changes

This is a tentative to open a discussion about a *development mode*  in radosgw.


jecluis:
@giubacc can we make this a PID record instead of an ADR? We can make it an ADR if/when we decide to move forward with this.

giubacc:
ok, let's do that way

https://github.com/aquarist-labs/s3gw/pull/782

PR #738, opened 2023-10-04
Dockerfile: Add filesystems:ceph:s3gw repomd.xml

tserong:
Without this, the subsequent `zypper install ; zypper ref` is effectively a no-op if docker build has already been run and cached that layer, which means we don't pick up updated package versions.

# Describe your changes


jecluis (approved):
this is probably one of the weirdest patches I've reviewed in the last year

PR #757, opened 2023-10-13
docs/ideas: define PID records

jecluis:
This patch introduces the concept of Project Idea Document (PID), meant to record ideas in a simple, straighforward way, for future reference and to track all the bright ideas we have.

These are meant to track ideas that

  - are not yet planned to be worked on
  - are not fully designed
  - may need to be evaluated at a later time

However, all PIDs should include enough information to assess their merit before being merged.

### DIFF
--- a/docs/ideas/PID-0-reserved.md
+++ b/docs/ideas/PID-0-reserved.md
@@ -1,0 +1,71 @@
+---
+status: accepted
+tags: area/docs
+updated: 20231013
+---
+
+# Project Idea Document (PID)
+
+## Context and Problem Statement
+
+We want to record ideas for features, enhancements, and reasonably sane brain
+dumps.
+
+Project Idea Documents (henceforth, PID) are therefore records, much like the
+ADRs we use for decisions, but fall on a different category: these are not
+decisions made by anyone, just a description of something to be considered for
+the future.
+
+## Format
+
+PIDs should observe the following qualities:
+
+- Easy to read and review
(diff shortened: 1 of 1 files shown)


irq0 (approved):

tserong (approved):

l-mb:
I don't understand what this brings over tracking such draft ideas to investigate further in our issue tracker?

jecluis:
> I don't understand what this brings over tracking such draft ideas to investigate further in our issue tracker?

I see your point. This might feel like we're tracking issues in documents. But I tend to see issues a bit more ephemeral, given that once they are closed/rejected they tend to be consigned to oblivion. So a few things come to mind:

- being persistent ad eternum, allowing them to be referenced if needed
- easier to move to an ADR
- issues in the tracker are expected to be worked on or eventually closed, whereas this would be a stable repository of things that are in neither state
- issues that get closed in the tracker tend to be forever forgotten; whereas things that get rejected here would still exist, not go away, and at some point someone could go like "oh, yeah, this makes sense"
- better for documentation, when new comers join, because no one who joins the project is going to go through past issues that have been resolved or rejected, whereas this can technically be documentation to read
- and issues don't get properly reviewed before being filed, whereas these would have to be reviewed and their merit assessed before being merged

votdev (approved):
LGTM and worth to give it a try

giubacc (approved):
why not, LGTM

0xavi0 (approved):
lg2m

PR #758, opened 2023-10-14
Dockerfile: use wrapper script instead of radosgw directly

### LINKED ISSUE #754: docs: explain how to use `--rgw-dns-name` to enable hostname style buckets
This is already handled by the charts for k8s deployments, but for dev/test or standalone usage, we should document somewhere to run with `--rgw-dns-name YOUR_DOMAIN_NAME` in case hostname style bucket access is desired.  For example, in the ephemeral docker case you'd run something like this:

`docker run -p 7480:7480 quay.io/s3gw/s3gw:latest --rgw-backend-store sfs --debug-rgw 1 --rgw-dns-name YOUR_DOMAIN_NAME`

(This was originally reported on Slack at https://rancher-users.slack.com/archives/C04DCMUV8SE/p1696951336295009 when testing using a Java app using Amazon's S3Client class. This class apparently defaults to hostname style buckets, but can be made to using path style buckets by setting forcePathStyle = true.)

jecluis:
This patch set introduces a wrapper script, `run-s3gw.sh`, responsible for abstracting calling the `radosgw` binary.

Instead, we define a set of options that are s3gw-specific and translate them to `radosgw` options as needed.
This ensures consistency in the project, by removing direct references to `radosgw`.

Additionally, we're adding documentation about advanced usage of the s3gw container, for cases when using the container directly may be needed.

This pull request has a companion pull request in aquarist-labs/s3gw-charts#127.

Fixes: aquarist-labs/s3gw#754

### DIFF
--- a/docs/advanced-usage.md
+++ b/docs/advanced-usage.md
@@ -1,0 +1,121 @@
+# Advanced usage
+
+The processes and steps described in this page should not be applicable to the
+vast majority of s3gw deployments. However, depending on circumstance, one might
+need to perform some of these for testing or debugging.
+
+The following sections rely on `podman` for examples. Using `docker` instead
+should work in the same way, although option names may need to be adjusted on
+occasion.
+
+## Running the container
+
+One might need to run the container standalone, in which case it's important to
+understand how to properly configure it. Running the container with `--help`
+will provide some insight.
+
+`podman run -it quay.io/s3gw/s3gw:latest --help`
+
+Which should result in something like the following, minus eventual changes not
+reflected in this example:
+
+    usage: /s3gw/bin/entrypoint.sh [options... -- args]
+
+    options:
+    --help                Shows this message.
+    --id ID               Specifies a custom instance ID (default: s3gw).
+    --cert FILE KEY       Specifies SSL certificate. Expects a .crt as first
+                          argument, and a .key as second argument.
+    --dns-name VALUE      For vhost style buckets, VALUE should be the DNS domain
+                          to be used.
+    --debug LEVEL         Runs with debug. Levels available: "high", "medium",
+                          "low", "none" (default: none).
+    --no-telemetry        Disable telemetry.
+    --telemetry-url URL   Specifies telemetry URL.
+    --with-status         Enables status frontend at port 9090.
+
+    args:
+    Any option supported by RADOS Gateway. For advanced use only.
+
+    env variables:
+    S3GW_ID           Specifies a custom instance ID.
+    S3GW_DNS_NAME     Specifies a DNS domain to be used for vhost style buckets.
+    S3GW_DEBUG        Specifies the debug level to be used.
+    S3GW_CERT_FILE    Specifies the SSL certificate file.
+    S3GW_CERT_KEY     Specifies the SSL certificate key file.
+
+Keep in mind that to access the service, one needs to expose the required ports.
+For s3gw these are port 7480 for non-SSL access, and port 7481 for SSL access.
+This can be achieved with:
+
+`podman run -it -p 7480:7480 quay.io/s3gw/s3gw:latest`
+
+In the following sections, we will be describing some of the available options.
+
+### Using vhost style buckets
+
+S3 supports two styles of bucket access: path and vhost. The former considers
+the buckets will be part of the address path, like in `my.domain.tld/my-bucket/`,
+whereas the latter assumes the bucket will be part of the vhost,
+`my-bucket.my.domain.tld`.
+
+By default, s3gw runs with path style buckets. One can enable vhost style
+buckets by providing the DNS domain name to be used for the service. For
+example,
+
+    podman run -it -p 7480:7480 \
+        quay.io/s3gw/s3gw:latest \
+        --dns-name my.domain.tld
+
+### Specifying certificates
+
+s3gw supports running with SSL, but will require a certificate to be provided.
+The location of this certificate must be local to the container namespace, and
+thus should be bind mounted into the container.
+
+    podman run -it -p 7481:7481 \
+        -v ./my-cert.crt:/data/my-cert.crt \
+        -v ./my-cert.key:/data/my-cert.key \
+        quay.io/s3gw/s3gw:latest \
+        --cert /data/my-cert.crt /data/my-cert.key
+
+The s3gw service should now be available on the host's port 7481 with SSL
+enabled.
+
+### Debugging output
+
+s3gw supports four different log levels: `none`, `low`, `medium`, and `high`. By
+default, s3gw runs with `none`.
+
+The vast majority of deployments will only require `none` or `low`, for as long
+as they are behaving nicely. In case of misbehavior, it is reasonable to look
+towards the logs for indication of what the problem might be. In this case, the
+typical deployment will rely on `medium`. Only in very exceptional situations
+will a `high` be useful. Keep in mind that at a `high` debug level, the system
+will be put under a lot of strain from the amount of outputted messages, and the
+value of additional verbosity may not be significant.
+
+Running with debug only requires specifying the `--debug LEVEL` argument to the
+container, such as
+
+    podman run -it -p 7480:7480 \
+        quay.io/s3gw/s3gw:latest \
+        --debug medium
+
+### Environment variables
+
+The available environment variables reflect the arguments accepted by the
+container. These can be provided to the container using the `--env VAR=VALUE`
+argument to `podman run`. For example,
+
+    podman run -it -p 7480:7480 \
+        --env S3GW_DEBUG=medium \
+        quay.io/s3gw/s3gw:latest
+
+### Additional arguments
+
+The container accepts additional arguments, that will be passed directly to the
+underlying RADOS Gateway binary. These should be used only if one knows what
+they are doing. To obtain a small subset of the option supported, running with
+`-- --help` will provide some information. Please note that not all the options
+provided in this help message will be applicable.
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,39 +5,39 @@ FROM registry.suse.com/bci/bci-base:15.5 as s3gw-base
 ADD https://download.opensuse.org/repositories/filesystems:/ceph:/s3gw/15.5/repodata/repomd.xml /tmp/repodata-s3gw.xml
 # Add OBS repository for additional dependencies necessary on Leap 15.5
 RUN zypper ar \
-  https://download.opensuse.org/repositories/filesystems:/ceph:/s3gw/15.5/ \
-  s3gw-deps \
- && zypper --gpg-auto-import-keys ref
+    https://download.opensuse.org/repositories/filesystems:/ceph:/s3gw/15.5/ \
+    s3gw-deps \
+    && zypper --gpg-auto-import-keys ref
 
 # Try `zypper install` up to three times to workaround mirror timeouts
 RUN for i in {1..3} ; do zypper -n install \
-  libsqlite3-0=3.43.1 \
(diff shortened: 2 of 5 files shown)


docs/advanced-usage.md
```diff
@@ -0,0 +1,118 @@
...
+    --id ID           Specifies a custom instance ID (default: s3gw).
+    --cert FILE KEY   Specifies SSL Certificate. Expects a .crt as first
+                      argument, and a .key as second argument.
+    --dns-name VALUE  For vhost style buckets, VALUE should be the DNS domain to
+                      be used.
+    --debug LEVEL     Runs with debug. Levels available: "high", "medium", "low",
+                      "none" (default: none)
+
+    args:
+    Any option supported by RADOS Gateway. For advance use only.
```
tserong:
```suggestion
    Any option supported by RADOS Gateway. For advanced use only.
```

tserong (commented):
I found one tiny typo :-) and the code looks fine, but I have two questions and one small problem. Here's the questions:

1. It looks like `--rgw-frontends` can't be overridden, which means there's no way to enable the status frontend. Should this PR also add `status` to frontend_args? Or will that be done separately?
2. Why support args _and_ env vars (as opposed to just one or the other)? Do we need to document that args will always override env vars?

The small problem is that since this change there's maybe something janky about signal handling. I ran `docker build --file Dockerfile --tag s3gw  --target s3gw .` to build the thing, then `docker run -p7480:7480  s3gw` to run it, and it _runs_ just fine.  But when I hit CTRL-C in the terminal, nothing happens. It just keeps running. I have to `docker kill` it from elsewhere (compare with `docker run quay.io/s3gw/s3gw:latest`, for which CTRL-C kills it just fine).

docs/advanced-usage.md
jecluis: oops

jecluis:
> I found one tiny typo :-) and the code looks fine, but I have two questions and one small problem. Here's the questions:
> 
>     1. It looks like `--rgw-frontends` can't be overridden, which means there's no way to enable the status frontend. Should this PR also add `status` to frontend_args? Or will that be done separately?

that could technically be done by providing the argument in the end, after `--`, but good shout, that should be an option to add.

>     2. Why support args _and_ env vars (as opposed to just one or the other)? Do we need to document that args will always override env vars?

Honestly, covering all bases. Some might be more inclined to rely on `--env FOO=bar` or something. Also, I suspect that it might be easier for k8s to use env vars.

> The small problem is that since this change there's maybe something janky about signal handling. I ran `docker build --file Dockerfile --tag s3gw --target s3gw .` to build the thing, then `docker run -p7480:7480 s3gw` to run it, and it _runs_ just fine. But when I hit CTRL-C in the terminal, nothing happens. It just keeps running. I have to `docker kill` it from elsewhere (compare with `docker run quay.io/s3gw/s3gw:latest`, for which CTRL-C kills it just fine).

I thought that would only work if using `-it` ? If not, maybe a sighandler on the script might suffice. I'll test that out and see how that goes.

docs/advanced-usage.md
```diff
@@ -0,0 +1,121 @@
...
+    --id ID               Specifies a custom instance ID (default: ${default_id}).
+    --cert FILE KEY       Specifies SSL certificate. Expects a .crt as first
+                          argument, and a .key as second argument.
+    --dns-name VALUE      For vhost style buckets, VALUE should be the DNS domain
+                          to be used.
+    --debug LEVEL         Runs with debug. Levels available: "high", "medium",
+                          "low", "none" (default: ${default_debug_level}).
+    --no-telemetry        Disable telemetry.
+    --telemetry-url URL   Specifies telemetry URL.
+    --with-status         Enables status frontend at 127.0.0.1:9999.
```
tserong:
```suggestion
    --with-status         Enables status frontend at 127.0.0.1:9090.
```

docs/advanced-usage.md
```diff
@@ -0,0 +1,121 @@
...
+`podman run -it quay.io/s3gw/s3gw:latest --help`
+
+Which should result in something like the following, minus eventual changes not
+reflected in this example:
+
+    usage: /s3gw/bin/run-s3gw.sh [options... -- args]
+
+    options:
+    --help                Shows this message.
+    --id ID               Specifies a custom instance ID (default: ${default_id}).
```
tserong:
```suggestion
    --id ID               Specifies a custom instance ID (default: s3gw).
```

docs/advanced-usage.md
```diff
@@ -0,0 +1,121 @@
...
+    --cert FILE KEY       Specifies SSL certificate. Expects a .crt as first
+                          argument, and a .key as second argument.
+    --dns-name VALUE      For vhost style buckets, VALUE should be the DNS domain
+                          to be used.
+    --debug LEVEL         Runs with debug. Levels available: "high", "medium",
+                          "low", "none" (default: ${default_debug_level}).
```
tserong:
```suggestion
                          "low", "none" (default: none).
```

tools/run-s3gw.sh
```diff
@@ -0,0 +1,209 @@
...
+  --id ID               Specifies a custom instance ID (default: ${default_id}).
+  --cert FILE KEY       Specifies SSL certificate. Expects a .crt as first
+                        argument, and a .key as second argument.
+  --dns-name VALUE      For vhost style buckets, VALUE should be the DNS domain
+                        to be used.
+  --debug LEVEL         Runs with debug. Levels available: "high", "medium",
+                        "low", "none" (default: ${default_debug_level}).
+  --no-telemetry        Disable telemetry.
+  --telemetry-url URL   Specifies telemetry URL.
+  --with-status         Enables status frontend at 127.0.0.1:9999.
```
tserong:
```suggestion
  --with-status         Enables status frontend at 127.0.0.1:9090.
```

tools/run-s3gw.sh
```diff
@@ -0,0 +1,209 @@
...
+  Options always override environment variables.
+EOF
+}
+
+stop_rgw() {
+  kill -n SIGTERM ${rgw_pid}
+  exit 1
+}
+
+trap "stop_rgw" EXIT
```
tserong: This needs to be moved down the bottom, just above when we run `radosgw` on line 199. Otherwise, if I try `docker run s3gw --help`, I get the help output, but I _also_ see "kill: usage: kill [-s sigspec | -n signum | -sigspec] pid | jobspec ... or kill -l [sigspec]" printed to STDERR. This is because `--help` triggers `exit 0` down on line 81, which hits the trap that calls `stop_rgw()`, which then tries to run `kill` with `$rgw_pid` as an empty string. Same with the various other exit cases. Moving `trap "stop_rgw" EXIT` down after those other `exit`s fixes this.

tserong (changes requested):
Good news! CTRL-C works to kill the container now when run from docker (I didn't try podman).

I've found a couple more typos, and one small kink (see review comments below).

Also, while `--with-status` does work to start the status backend (I see "starting handler: status in the logs) I can't connect to the damn thing when running `docker run -d -p7480:7480  -p9090:9090  s3gw  --with-status`. I just get "connection reset" if I try to hit http://localhost:9090 in my browser. This seems to be because the status backend binds to 127.0.0.1 inside the container by default. If I do this:

```diff 
 if [[ $with_status -eq 1 ]]; then
-  frontend_args+=", status"
+  frontend_args+=", status bind=0.0.0.0"
 fi
```

...it works fine. But I'm not sure we should actually make that change here in the runner script. Presumably that should be done in rgw_status_frontend.cc (why _does_ that bind to 127.0.0.1 inside the container? AFAICT that makes in inaccessible to the outside)

jecluis:
> ...it works fine. But I'm not sure we should actually make that change here in the runner script. Presumably that should be done in rgw_status_frontend.cc (why _does_ that bind to 127.0.0.1 inside the container? AFAICT that makes in inaccessible to the outside)

That's a good question. I always thought it could be exposed even if bound to 127.0.0.1. It's a shame it isn't. I think the rationale is that said sort of information should be consumed by _something_, not necessarily _someone_. If we expose it in  0.0.0.0, then we'll likely have to add authentication to it, whereas running on 127.0.0.1 would allow the admin to do shenanigans to get to it if they want to.

In principle, that was it. Given it's not reaching the port from outside the container though, I think the wrapper script needs to handle that. I just think there's no good way around it. Changing the status frontend to bind to 0.0.0.0 defeats the locality purpose of that frontend, and forces us to implement authentication for it.

tserong:
I found https://stackoverflow.com/questions/52513336/is-there-a-way-to-expose-a-docker-container-port-bound-to-127-0-0-1-to-host#52518929 on stackoverflow which is interesting.

Anyway, I think what you've got here now with `status bind=0.0.0.0` in the wrapper is fine, and does what we want, because the admin/user has explicitly decided to run it `--with-status`. And given what you've said, it probably makes sense to leave the default bind as 127.0.0.1 in the backend code, because that could theoretically be built by someone and run in other non containerized environments. And we can re-visit that decision if necessary later in a separate issue/PR/whatever.

tserong (dismissed):

jecluis:
merge dependent on me following up on aquarist-labs/s3gw-charts#127 -- these should be merged in tandem so we don't break the chart.

tools/run-s3gw.sh
```diff
@@ -0,0 +1,209 @@
...
+  --id ID               Specifies a custom instance ID (default: ${default_id}).
+  --cert FILE KEY       Specifies SSL certificate. Expects a .crt as first
+                        argument, and a .key as second argument.
+  --dns-name VALUE      For vhost style buckets, VALUE should be the DNS domain
+                        to be used.
+  --debug LEVEL         Runs with debug. Levels available: "high", "medium",
+                        "low", "none" (default: ${default_debug_level}).
+  --no-telemetry        Disable telemetry.
+  --telemetry-url URL   Specifies telemetry URL.
+  --with-status         Enables status frontend at port 9090.
```
m-ildefons: The docker file says `EXPOSE 7481`, not `9090`

m-ildefons (dismissed):
LGTM.
It's common practice to name an entrypoint script `entrypoint.sh`

tools/run-s3gw.sh
jecluis: 7481, I think, is the SSL port. I didn't add the 9090 to the dockerfile, wonder if I should even if that's merely for documentation purposes.

irq0:
Heads up. Some more dependencies:

https://github.com/aquarist-labs/s3gw/blob/37bd64e3a277d22c190f6048e59b6d9303ea6dc4/tools/s3tests/runner.py#L58

https://github.com/irq0/vogon2/blob/5772347d8b8dff4c6da8a103402d1f39750e5116/vogon2.py#L130

s3tr needs changes: It doesn't use entrypoint, but command

vogon needs some changes ✔️

jecluis:
> Heads up. Some more dependencies:
> 
> https://github.com/aquarist-labs/s3gw/blob/37bd64e3a277d22c190f6048e59b6d9303ea6dc4/tools/s3tests/runner.py#L58
> 
> https://github.com/irq0/vogon2/blob/5772347d8b8dff4c6da8a103402d1f39750e5116/vogon2.py#L130
> 
> s3tr needs changes: It doesn't use entrypoint, but command
> 
> vogon needs some changes ✔️

@irq0 will any of these fail because of these changes? Looking at them it seems that they rely on executing the binary directly, not the entrypoint?

jecluis:
@m-ildefons @tserong rebased and addressed the comments from before. Mind another look?

tserong (approved):
LGTM (I gave it a quick test too, both with and without `--with-status`)

Issue #749, opened 2023-10-06
[FR] Download multipart parts


0xavi0:
This was discovered when running the `warp multipart` test as it always complains about the size.
`warp` uploads parts and then tries to download them. (not the whole object)

When getting multiparts parts we're returning the whole object.

Example:
```bash
// create a dummy part (we'll use the same for every part upload)
$ dd if=/dev/random of=/tmp/part bs=1M count=6

// create the multipart upload (take note of the uploadId)
$ aws --endpoint=http://127.0.0.1:7480 s3api create-multipart-upload  --bucket test --key test-mp


// upload parts
$ aws --endpoint=http://127.0.0.1:7480 s3api upload-part --bucket test --key test-mp --body /tmp/part --upload-id 20231006T151146.621816566Z --part-number 1
$ aws --endpoint=http://127.0.0.1:7480 s3api upload-part --bucket test --key test-mp --body /tmp/part --upload-id 20231006T151146.621816566Z --part-number 2
$ aws --endpoint=http://127.0.0.1:7480 s3api upload-part --bucket test --key test-mp --body /tmp/part --upload-id 20231006T151146.621816566Z --part-number 3
$ aws --endpoint=http://127.0.0.1:7480 s3api upload-part --bucket test --key test-mp --body /tmp/part --upload-id 20231006T151146.621816566Z --part-number 4
$ aws --endpoint=http://127.0.0.1:7480 s3api upload-part --bucket test --key test-mp --body /tmp/part --upload-id 20231006T151146.621816566Z --part-number 5
$ aws --endpoint=http://127.0.0.1:7480 s3api upload-part --bucket test --key test-mp --body /tmp/part --upload-id 20231006T151146.621816566Z --part-number 6
$ aws --endpoint=http://127.0.0.1:7480 s3api upload-part --bucket test --key test-mp --body /tmp/part --upload-id 20231006T151146.621816566Z --part-number 7
$ aws --endpoint=http://127.0.0.1:7480 s3api upload-part --bucket test --key test-mp --body /tmp/part --upload-id 20231006T151146.621816566Z --part-number 8
$ aws --endpoint=http://127.0.0.1:7480 s3api upload-part --bucket test --key test-mp --body /tmp/part --upload-id 20231006T151146.621816566Z --part-number 9
$ aws --endpoint=http://127.0.0.1:7480 s3api upload-part --bucket test --key test-mp --body /tmp/part --upload-id 20231006T151146.621816566Z --part-number 10

[...]

PR #759, opened 2023-10-16
docs/matrix: update compatibility matrix

jecluis:
We don't support obtaining an object's specific part once the object's multipart upload has been finished.

This is a limitation of the current RGW implementation.

Also, update the matrix stating that we don't support bucket version suspension, nor do we support (for now) transitioning from an unversioned bucket to a versioned bucket.

References: aquarist-labs/s3gw#749


jecluis:
closing because no longer up to date

PR #760, opened 2023-10-16
s3tr: Ensure we print the important analyze summary info

irq0:
Use soft wrap and min width settings to ensure that the important info on s3 test results end up printed in full - not truncated.


m-ildefons (approved):
LGTM

PR #774, opened 2023-10-24
Dockerfile: retry `zypper install` up to three times

tserong:
We've variously had problems with `zypper install` timing out both in CI, and sometimes on local developer systems.  When this happens it kills the entire build, which is extremely irritating.  It seems we're not the first project to hit this problem (see https://github.com/openSUSE/zypper/issues/420).

My hope here is that by retrying `zypper install` if the first call fails due to a dodgy mirror (or whatever), the second or third call will succeed.  At best this amounts to a no-op.  At worst, it'll completely fail more slowly than it did before.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,12 +3,14 @@ FROM opensuse/leap:15.4 as s3gw-base
 # This makes sure the Docker cache is invalidated
 # if packages in the s3gw repo on OBS have changed
 ADD https://download.opensuse.org/repositories/filesystems:/ceph:/s3gw/15.4/repodata/repomd.xml /tmp/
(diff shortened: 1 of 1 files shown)


giubacc:
I'm not against this, just it could be worth to put a comment in the Dockerfile to avoid the "what the heck" surprise of the casual reader.

jecluis (commented):
Agree with @giubacc - a comment in the beginning of the dockerfile or something explaining what we are doing seems reasonable. Other than that, lgtm.

tserong:
Thanks guys, done. I've also moved another (unrelated) comment to the correct location.

jecluis (approved):

giubacc (approved):

PR #782, opened 2023-10-26
PID - radosgw development mode

giubacc:
# Describe your changes

PID about an hypothesis for a development mode in radosgw.

### DIFF
--- a/docs/ideas/PID-20231026-radosgw-dev-mode.md
+++ b/docs/ideas/PID-20231026-radosgw-dev-mode.md
@@ -1,0 +1,39 @@
+---
+status: proposed
+tags: radosgw
+updated: 20231026
+---
+
+# radosgw development mode
+
+## Context and Problem Statement
+
+For development/testing purposes it could be worth having the ability to enable
+a so called: *development mode* in the `radosgw`.
+Such mode should obviously be disabled by default and should be explicitly
+enabled with a dedicated parameter set by the developer/tester.
+When under the *development mode*, the `radosgw` would accept certain kind of requests
+otherwise ignored.
+
+Decide on the merits of a single request type is not in the scope of this ADR, here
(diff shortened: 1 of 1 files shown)


docs/ideas/PID-20231026-radosgw-dev-mode.md
```diff
@@ -0,0 +1,39 @@
...
+we want to establish whether the presence of the development mode is admissible
+per se.
+
+It follows a list of request categories that could be enabled when the `radosgw`
+is under the *development mode*:
+
+- Fault injections
+- Artificial slowdowns
+- Event dispatching to other development/testing components
+
```
jecluis, line 28: I think it would be important to describe, even if briefly, how these would be communicated to rgw.

Are we talking a new frontend with a REST API? Or just parameters to the binary?

giubacc: The initial purpose of this idea was to establish if we consider admissible a general `dev mode` avoiding to descend too much in specific details.
So my first question is: do we need a `dev mode`? Could it be useful to us?
In the future we could have the need to inject something such as: a failure, a slowdown, to observe the behavior of the radosgw in that circumstance; say precisely, now, what that could be, I honestly dunno, it is just a need that I think could be possible in the future.
The tentative of this PID is to give a hook, a chance, to us, if there will be that need in the future, to allow a dev related feature to be part of the official radosgw codebase and not to develop that in a manner that won't be ever allowed to be merged because, without a dev mode theorized, it would be clearly something not mergeable.
To respond more precisely in the merit of your question, I'd prefer a REST API approach to trigger specific dev behaviors, that would allow those things to be turned on/off, configured, when the process is running and not statically defined at the start.
This would also allow sending back, via REST calls issued by the radosgw, eventual unsolicited responses containing information/events regarding the "dev thing" that has been triggered.

Issue #771, opened 2023-10-23
Update container base image


l-mb:
Currently, the s3gw container is built with leap:15.4 as the base image. This goes end of maintenance in November 2023.

s3gw should at least rebase to leap:15.5, or even consider utilizing the same base image as longhorn (https://github.com/longhorn/longhorn-instance-manager/blob/f80354ec9a09e986f378017410891b26dddecc00/package/Dockerfile#L1), which uses the sle-bci.

tserong:
I've just tried doing this:

```diff
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,11 @@
-FROM opensuse/leap:15.4 as s3gw-base
+FROM registry.suse.com/bci/bci-base:15.5 AS s3gw-base
 
 # This makes sure the Docker cache is invalidated
 # if packages in the s3gw repo on OBS have changed
-ADD https://download.opensuse.org/repositories/filesystems:/ceph:/s3gw/15.4/repodata/repomd.xml /tmp/
-# Add OBS repository for additional dependencies necessary on Leap 15.4
+ADD https://download.opensuse.org/repositories/filesystems:/ceph:/s3gw/SLE_15_SP5/repodata/repomd.xml /tmp/
+# Add OBS repository for additional dependencies necessary on SLE 15 SP5
 RUN zypper ar \
-  https://download.opensuse.org/repositories/filesystems:/ceph:/s3gw/15.4/ \
+  https://download.opensuse.org/repositories/filesystems:/ceph:/s3gw/SLE_15_SP5/ \
   s3gw-deps \
  && zypper --gpg-auto-import-keys ref
```

...which works fine to build the s3gw-base image. But then when we get on to the buildenv image, there's a handful of dependencies that aren't available in BCI:

```
No provider of 'gmock' found.
No provider of 'gtest' found.
No provider of 'librabbitmq-devel' found.
No provider of 'memory-constraints' found.
No provider of 're2-devel' found.
No provider of 'xmlstarlet' found.
Package 'lua53-luarocks' not found.
```

Some of these I can presumably get from some other devel repo on OBS, or I can just link them in to the filesystems:ceph:s3gw project. Also, this raises the question of whether we can drop any of these. For example, we don't build rgw's AMQP endpoint so presumably don't need librabbitmq-devel in the first place.

l-mb:
[...]

PR #786, opened 2023-10-30
Update container base image to leap or bci 15.5

tserong:
There's two commits here for review. Depending on which way we want to go we can either just take the first one and get leap 15.5, or I can squash the two to give us bci 15.5.

The leap 15.5 change is extremely straightforward (`s/15.5/15.5/g`). With bci though, we lack a bunch of ceph's build dependencies. Rather that drag all those over into filesystems:ceph:s3gw on OBS (which was looking to be a PITA), I've added the _leap_ 15.5 repos to the buildenv image. AIUI this is safe because 15.5 should be binary compatible across the board, and anyway it only affects the build environment, not s3gw-base or the final s3gw container. Does this sound sane to everyone?

I haven't removed any other build deps as mentioned in #771, as I wanted to focus of setting the base image first. I'll do more stripping if I can later.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,11 @@
-FROM opensuse/leap:15.4 as s3gw-base
+FROM registry.suse.com/bci/bci-base:15.5 as s3gw-base
 
 # This makes sure the Docker cache is invalidated
 # if packages in the s3gw repo on OBS have changed
-ADD https://download.opensuse.org/repositories/filesystems:/ceph:/s3gw/15.4/repodata/repomd.xml /tmp/
-# Add OBS repository for additional dependencies necessary on Leap 15.4
+ADD https://download.opensuse.org/repositories/filesystems:/ceph:/s3gw/15.5/repodata/repomd.xml /tmp/repodata-s3gw.xml
+# Add OBS repository for additional dependencies necessary on Leap 15.5
 RUN zypper ar \
-  https://download.opensuse.org/repositories/filesystems:/ceph:/s3gw/15.4/ \
+  https://download.opensuse.org/repositories/filesystems:/ceph:/s3gw/15.5/ \
   s3gw-deps \
  && zypper --gpg-auto-import-keys ref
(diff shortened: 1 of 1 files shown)


tserong:
> AIUI this is safe because 15.5 should be binary compatible across the board, and anyway it only affects the build environment, not s3gw-base or the final s3gw container. Does this sound sane to everyone?

Hang on, that might sound sane, but it's not _quite_ right. I've just got a test build where the radosgw binary is linking against liblua5.4.so.5 but the s3gw container only has liblua5.3.so.5. >sigh<

The _easy_ fix for that would presumably be to add those leap update repos to s3gw-base, but I'm not sure that's the _right_ fix.

tserong:
> The _easy_ fix for that would presumably be to add those leap update repos to s3gw-base, but I'm not sure that's the _right_ fix.

No, that's not the right fix.  The right fix is to specifically install lua53-devel, not lua-devel.

But after that I seem to have problems with gtest being slightly newer in 15.4 backports (1.12.1) than what's available in 15.5 anywhere (1.10.0), so the build fails because `GTEST_FLAG_SET` doesn't exist.

tserong:
> But after that I seem to have problems with gtest being slightly newer in 15.4 backports (1.12.1) than what's available in 15.5 anywhere (1.10.0), so the build fails because `GTEST_FLAG_SET` doesn't exist.

Linking googletest from Factory into filesystems:ceph:s3gw on OBS fixes that.

tserong:
OK, this should be good now. I got a test build via https://github.com/aquarist-labs/ceph/pull/242

jecluis (approved):
lgtm

tserong:
Thanks @jecluis. I've squashed this back to one commit now given I got it working with BCI. Any objections to this @m-ildefons?

jecluis:
@m-ildefons ping

m-ildefons (approved):
lgtm

PR #793, opened 2023-11-03
readthedocs: specify build.os

jecluis:
Builds are failing for a few weeks, due to a missing build.os.


m-ildefons (approved):
I have no idea if this fixes things or breaks things even further, but we should give it a try.

PR #804, opened 2023-11-17
Dockerfile: allow specifying s3gw directory

jecluis:
When requiring the copy of the `entrypoint.sh` script, we did not account for actions on other repositories relying on checking out the `s3gw.git` repository and building images using the Dockerfile from within an `s3gw` directory. This causes the script to not be found, because it's instead in `s3gw/tools/entrypoint.sh`, versus what we expected, `tools/entrypoint.sh`.

This helps addressing that by allowing the caller to specify on which directory we can find the s3gw repository.

Signed-off-by: Joao Eduardo Luis <joao@suse.com>


m-ildefons (approved):
Approved.

PR #813, opened 2023-11-20
Fix release workflow

m-ildefons:
# Describe your changes

Drop docker buildx from the release workflow. The builds using buildx had various issues with timeouts and TLS errors.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -16,18 +16,6 @@ jobs:
         with:
           submodules: recursive
 
-      - name: Set up QEMU
-        uses: docker/setup-qemu-action@v2
-
-      - name: Set up Docker Context
-        run: |
-          docker context create builder
-
-      - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v2
-        with:
-          endpoint: builder
-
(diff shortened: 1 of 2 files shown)


m-ildefons:
The release workflow is expected to fail on this step since there are no release notes matching the dummy release tag provided. However, the build of the container succeeded, which is the important part.

jecluis (dismissed):
:partying_face:

jecluis (dismissed):

.github/workflows/release.yaml
```diff
@@ -6,6 +6,25 @@ on:
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Setup Python
+        uses: actions/setup-python@v3
+
+      - name: Get changed files
+        id: changed
+        uses: tj-actions/changed-files@v40
```
jecluis: I fear this may cause the same problem we were seeing with the other branches. Any chance we can validate it before merging to main?

m-ildefons: I'll just drop this then. I just want the release pipeline fixed.

jecluis (approved):

